PR #3, opened 2020-01-22
Initial setup for containerizing code

lazypanda10117:
As stated.


danielchen115:
Is there any set up we need to do on our local machines or any changes to how we run our containers?

PR #5, opened 2020-01-28
Initialize sequelize and add migrations, seeds, resolvers/schema for users/events

danielchen115:
## Overview

[Notion task](https://www.notion.so/uwblueprintexecs/DB-setup-bdf94d548b8d471c9b83ef29fa4374d4)

GraphQL queries for events and users that hit the database for `events`, `event(id: id)`, `users`, `user(id: id)`. Mutations are not supported yet.

## Changes
Initialize ORM and graphql connectors. Add models, migrations, seeders, schemas, resolvers for events and users.

## Screenshots
Example usage:
image

image

## Testing
Manual testing as per the screenshots. Play with it and try to break it (won't be hard for now :P, not a lot of functionality yet)!

## Other Notes 
**Setup:** 
1. Kill and restart containers to install new node packages.
2. From the node container shell (command in wiki), run:
        `sequelize db:migrate`
        `sequelize db:seed`

### DIFF
--- a/schema/user.js
+++ b/schema/user.js
@@ -5,6 +5,7 @@ const userSchema = `
   }
   
   type User {
+    id: Int
     firstName: String
     lastName: String
     email: String
--- a/schema/user.js
+++ b/schema/user.js
@@ -1,0 +1,15 @@
+const userSchema = `
+  extend type Query {
+    user(id: Int!): User
+    users: [User]
+  }
+  
+  type User {
+    firstName: String
+    lastName: String
+    email: String
+  }
+  `
+;
+
+exports.userSchema = userSchema;
--- a/schema/user.js
+++ b/schema/user.js
@@ -5,7 +5,7 @@ const userSchema = `
   }
   
   type User {
-    id: Int
+    id: ID!
     firstName: String
     lastName: String
     email: String
--- a/seeders/20200128023056-demo-user.js
+++ b/seeders/20200128023056-demo-user.js
@@ -2,13 +2,29 @@
 
 module.exports = {
   up: (queryInterface, Sequelize) => {
-    return queryInterface.bulkInsert('users', [{
-      firstName: 'John',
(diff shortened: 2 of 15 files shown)


schema/user.js
```diff
@@ -0,0 +1,16 @@
+const userSchema = `
```
petorsong: should events also have a resolver + schema?

server.js
```diff
@@ -1,32 +1,40 @@
...
+const { userResolvers } = require('./resolvers/user');
+
 const express = require('express');
 const bodyParser = require('body-parser');
 const { graphqlExpress, graphiqlExpress } = require('apollo-server-express');
 const { makeExecutableSchema } = require('graphql-tools');
 
+const { Sequelize } = require('sequelize');
+
+// TODO: this stuff can be set to env variables
```
petorsong: we should set this up in this PR, it's hardcoded in too many places right now

seeders/20200128023056-demo-user.js
```diff
@@ -0,0 +1,32 @@
+'use strict';
+
+module.exports = {
+  up: (queryInterface, Sequelize) => {
+    return queryInterface.bulkInsert('users', [
```
petorsong: what's the difference between a seeder and inserting an entry via migration?

schema/user.js
danielchen115: yep! just did the user one first to see how it works. I can do the events one tonight

seeders/20200128023056-demo-user.js
danielchen115: We may have integrity issues if we have migrations that modify tables where previous migrations already seeded data. If we separate them, we can re-seed freely without having to redo single migrations (which can be dangerous if there are more recent dependent migrations)

Also on prod/test envs, we might need to migrate without seeding

PR #9, opened 2020-02-04
Add new event columns

JasonYG:
## Overview
Add date, createdBy, and isActive columns to the events table.
[Notion task](https://www.notion.so/uwblueprintexecs/events-db-setup-f9c6173178f447968d667e7bb2f16f2c)

## Changes
- updated the events schema to include new columns
- created new migration to add and delete columns
- updated events model to include new columns
- updated seed to include new dummy data

## Screenshots
image

## Testing
- Manual testing on graphiql


danielchen115:
Nice work 💯 ! The createdBy field needs to be a _many-to-one_ foreign key to the users table. 
[helpful link](https://sequelize.org/master/manual/assocs.html)
Also, the schema and resolver needs to reflect that as well. [helpful link](https://stackoverflow.com/questions/53905806/understanding-include-when-resolving-source-association-from-target-in-sequeli)

No worries at all if you don't get around to it! I can do it later this week.

JasonYG:
@danielchen115 I've updated the schema and model to reflect the many-to-one relationship of events to users. Could you take a look at these updates, as well as how I can create seeders? Thanks!

models/event.js
```diff
@@ -1,12 +1,21 @@
...
-    pin: DataTypes.STRING,
-    description: DataTypes.TEXT
-  }, {});
-  Event.associate = function(models) {
-    // associations can be defined here
+  const Event = sequelize.define(
+    "event",
+    {
+      name: DataTypes.STRING,
+      date: DataTypes.STRING,
```
lazypanda10117: Just wondering, if this means the date of an event, can there be event that spans multiple days?

danielchen115:
> @danielchen115 I've updated the schema and model to reflect the many-to-one relationship of events to users. Could you take a look at these updates, as well as how I can create seeders? Thanks!

Lgtm! For seeding, we can create associated users in the event seeder.

PR #11, opened 2020-02-04
Add date, time, datetime custom scalars

danielchen115:
## Overview
Allow for date/time/datetime data types when defining schema.

[Notion task](https://www.notion.so/uwblueprintexecs/Add-Date-scalar-graphql-997ef0b0c9cd40c58052ecd26c8f5a1f)

## Changes
* Installed graphql-iso-date
* imported scalars when creating schemas
* add example usage for users

## Screenshots
In the schema, define as so:
image

## Testing
manual testing on graphiql

## Other Notes
* run `npm install` for graphql-iso-date
* I put the schema updates in server.js. At some point, we should move the graphql stuff into another file.

@uwblueprint/paramedics

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -811,6 +811,11 @@
         "source-map-support": "^0.5.1"
       }
     },
+    "graphql-iso-date": {
+      "version": "3.6.1",
+      "resolved": "https://registry.npmjs.org/graphql-iso-date/-/graphql-iso-date-3.6.1.tgz",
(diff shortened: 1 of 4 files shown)


JasonYG:
> Do we need to run npm install for the new library? Shouldn't this already be in the container?

The library is stored in node_modules, meaning that we have to install the new dependency on our local working environment (since we don't commit node modules).

PR #26, opened 2020-03-11
CRUD mutations for event table

JasonYG:
## Overview
Adding CRUD mutations to the events table.

Note:  Undo all migrations, then remigrate before running any code.

First:
`sequelize-cli db:migrate:undo:all`

Then: 
`sequelize-cli db:migrate`

[Notion task](https://www.notion.so/uwblueprintexecs/CRUD-operations-for-events-132c40bbac0a48918a374f170e9becf6)

## Changes
* Add create, update, and delete mutations to the events table 
    * Schema updated in `schema/event.js`
    * Resolvers updated in `resolvers/event.js`
* Linked changes in `graphql.js`
* Implented create, update and deletion resolvers in `schema/event.js` and `resolvers/event.js`
    * Note: The return type of resolvers is an object of the dataset itself. 
* Error handling for updates (suggesting that we make some error file to pull in common error outputs)
* Changed schema to reflect inbuilt Sequelize function error handling:
    * db.delete: Returns 1 if record deleted, 0 if deleted
    * Update resolvers: Returning Event object if object has been successfully updates; otherwise, we will throw a custom error ("Invalid event ID" or "Invalid user ID")

## Screenshots
Add event mutation
image

Update all event fields mutation
image

Update a few event fields mutation
image

Error handling for updating invalid events
image

Error handling for updating events with an invalid user
image

Output for deletion of existing events
image

Output for deletion of non-existing events
image

## Testing
* Manual testing via GraphIQL

## Checklist
- [x] Change title to be more descriptive (shouldn't be just branch name)
- [x] Update `Notion task` link
- [x] `Overview`
- [x ] \(Optional) Add inline comments in diff
- [x] \(Optional) `Screenshots`
- [x] \(Optional) `Testing` (N/A if nothing changed)
- [x] Request at least 2 reviewers
- [x] Assign yourself to the PR
- [x ] Make sure build passes

@uwblueprint/paramedics

### DIFF
--- a/migrations/20200213004252-change-date-column-names.js
+++ b/migrations/20200213004252-change-date-column-names.js
@@ -12,7 +12,7 @@ module.exports = {
           },
           { transaction: t }
         ),
-        queryInterface.changeColumn("events", "data", {
+        queryInterface.changeColumn("events", "date", {
           type: Sequelize.DATE,
           allowNull: false
         }),
--- a/migrations/20200213004252-change-date-column-names.js
+++ b/migrations/20200213004252-change-date-column-names.js
@@ -7,7 +7,7 @@ module.exports = {
         queryInterface.removeColumn("events", "description"),
         queryInterface.removeColumn("events", "date"),
         queryInterface.addColumn("events", "eventDate", {
-          type: Sequelize.DataTypes.DATE,
+          type: Sequelize.DataTypes.DATEONLY,
           allowNull: false
         })
       ]);
--- a/migrations/20200213004252-change-date-column-names.js
+++ b/migrations/20200213004252-change-date-column-names.js
@@ -1,0 +1,38 @@
+"use strict";
+
+module.exports = {
+  up: (queryInterface, Sequelize) => {
+    return queryInterface.sequelize.transaction(t => {
+      return Promise.all([
+        queryInterface.removeColumn(
+          "events",
+          "description",
+          {
+            type: Sequelize.DataTypes.STRING
+          },
+          { transaction: t }
+        ),
+        queryInterface.changeColumn("events", "data", {
+          type: Sequelize.DATE,
+          allowNull: false
+        }),
+        queryInterface.renameColumn("events", "date", "eventDate")
+      ]);
+    });
+  },
+  down: (queryInterface, Sequelize) => {
+    return queryInterface.sequelize.transaction(t => {
+      return Promise.all([
+        queryInterface.addColumn(
+          "events",
+          "description",
+          {
+            type: Sequelize.DataTypes.STRING
+          },
+          { transaction: t }
+        ),
+        queryInterface.renameColumn("events", "eventDate", "date")
+      ]);
+    });
+  }
+};
--- a/migrations/20200213004252-change-date-column-names.js
+++ b/migrations/20200213004252-change-date-column-names.js
@@ -4,35 +4,24 @@ module.exports = {
   up: (queryInterface, Sequelize) => {
     return queryInterface.sequelize.transaction(t => {
       return Promise.all([
-        queryInterface.removeColumn(
-          "events",
-          "description",
-          {
-            type: Sequelize.DataTypes.STRING
-          },
-          { transaction: t }
-        ),
-        queryInterface.changeColumn("events", "date", {
-          type: Sequelize.DATE,
+        queryInterface.removeColumn("events", "description"),
+        queryInterface.removeColumn("events", "date"),
+        queryInterface.addColumn("events", "eventDate", {
+          type: Sequelize.DataTypes.DATE,
           allowNull: false
-        }),
-        queryInterface.renameColumn("events", "date", "eventDate")
+        })
       ]);
     });
   },
   down: (queryInterface, Sequelize) => {
-    return queryInterface.sequelize.transaction(t => {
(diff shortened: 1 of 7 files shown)


resolvers/event.js
```diff
@@ -9,6 +9,53 @@ const eventResolvers = {
...
+    },
+    updateEvent: async (parent, args) => {
+      const event = await db.event.findByPk(args.id);
+      const user = args.createdBy
+        ? await db.user.findByPk(args.createdBy)
+        : true;
+      if (!(event && user)) {
+        if (!event) throw new Error("Invalid event ID");
+        else throw new Error("Invalid user ID");
+      }
```
petorsong: seems excessively complicated
```suggestion
      if (!event) {
        throw new Error("Invalid event ID");
      } else if (!user) {
        throw new Error("Invalid user ID");
      }
```

resolvers/event.js
```diff
@@ -9,6 +9,53 @@ const eventResolvers = {
...
+        eventDate: args.eventDate,
+        createdBy: args.createdBy,
+        isActive: args.isActive
+      });
+    },
+    updateEvent: async (parent, args) => {
+      const event = await db.event.findByPk(args.id);
+      const user = args.createdBy
+        ? await db.user.findByPk(args.createdBy)
+        : true;
```
petorsong: is this ternary necessary? more specifically should `args.createdBy` ever be falsey?

migrations/20200213004252-change-date-column-names.js
```diff
@@ -0,0 +1,27 @@
+"use strict";
+
+module.exports = {
+  up: (queryInterface, Sequelize) => {
+    return queryInterface.sequelize.transaction(t => {
+      return Promise.all([
+        queryInterface.removeColumn("events", "description"),
+        queryInterface.removeColumn("events", "date"),
+        queryInterface.addColumn("events", "eventDate", {
+          type: Sequelize.DataTypes.DATE,
```
petorsong: would prefer `DATEONLY` over `DATE` for the column type

migrations/20200213004252-change-date-column-names.js
```diff
@@ -0,0 +1,27 @@
+"use strict";
+
+module.exports = {
+  up: (queryInterface, Sequelize) => {
+    return queryInterface.sequelize.transaction(t => {
+      return Promise.all([
```
petorsong: i ran into an issue when running this migration - because sequelize doesn't guarantee migrations occur in a transaction, it's possible to put your db in a bad intermediate state
- for now, undoing and re-running all migrations solves the problem, but this won't be a good idea down the line
- for reference, two potential solutions are:
1. migrate the data so new non-null columns won't break the migration (fixing the root problem)
2. have every migration have only 1 change (blanket solution for all schema changes)

package.json
```diff
@@ -20,6 +20,7 @@
     "graphql-iso-date": "^3.6.1",
     "graphql-tools": "^4.0.6",
     "lodash": "^4.17.15",
+    "mysql2": "^2.1.0",
```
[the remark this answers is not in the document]
JasonYG: For sure - I had it there to run my code on prod.

resolvers/event.js
JasonYG: if somebody tries to update an event but doesn't pass in the `createdBy` field (i.e. they're not updating the createdBy field), then `args.createdBy` would be undefined (falsey). I'm not sure if this is the intended functionality - I just bypassed the problem by setting user to `true` when no createdBy argument is passed.

migrations/20200213004252-change-date-column-names.js
JasonYG: when I'm doing this change, should I create a new migration? Or can I update the existing one?

migrations/20200213004252-change-date-column-names.js
JasonYG: When I'm fixing these migrations, should I be updating this file? Or should I be creating a new migration?

resolvers/event.js
petorsong: gotcha, in that case something like this would be better - it's generally bad practice for a variable to possibly be two different types
```suggestion
      if (args.createdBy) {
        const user = await db.user.findByPk(args.createdBy);
        if (!user) {
          throw new Error("Invalid user ID");
        }
      }
```

migrations/20200213004252-change-date-column-names.js
petorsong: since we're already asking people to re-run migrations for this case, you can make the update to this existing migration - this will not be the correct pattern moving forward though

migrations/20200213004252-change-date-column-names.js
petorsong: see comment below, tldr just update this file for this PR

resolvers/event.js
```diff
@@ -9,6 +9,57 @@ const eventResolvers = {
...
+        const user = await db.user.findByPk(args.createdBy);
+        if (!user) {
+          throw new Error("Invalid user ID");
+        }
+      }
+      if (!event) {
+        throw new Error("Invalid event ID");
+      } else if (!user) {
+        throw new Error("Invalid user ID");
+      }
```
petorsong: this block is not necessary anymore

PR #27, opened 2020-03-15
Fix container breaking on prod

danielchen115:
## Overview
Fixed bug where starting the containers on prod mode causes an error about installing mysql2.

[Notion task](https://www.notion.so/uwblueprintexecs/Prod-is-broken-43cce1dbfed24c75b32ae637a30a081f)

## Changes
- Expose the port 4000 instead of 8080.
- Fix config.json to reflect the right db fields

## Testing
manual testing

## Other Notes
- I pointed the db to the dev db for now, we can change it once we get prod up and running
- interactive graphql won't work in prod by design. it'll say `GET query missing.` instead

@uwblueprint/paramedics

### DIFF
--- a/config/config.json
+++ b/config/config.json
@@ -17,11 +17,12 @@
     "operatorsAliases": false
   },
   "production": {
-    "username": "root",
-    "password": null,
-    "database": "database_production",
-    "host": "127.0.0.1",
-    "dialect": "mysql",
+    "username": "robot",
+    "password": "robot_pwd",
+    "database": "paramedics-db",
+    "host": "paramedics-db",
+    "port": 5432,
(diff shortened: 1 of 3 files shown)


config/config.json
```diff
@@ -17,11 +17,12 @@
     "operatorsAliases": false
   },
   "production": {
```
petorsong: can we add a `TODO` here just as a reminder to change this to actual prod values once we have that set up

danielchen115: Hmm apparently you can't add comments to json files... can we make a note on our notion task board instead?

petorsong: whoops my bad, i'll put it on a ticket instead

PR #46, opened 2020-07-07
Patient schema changes

aaronabraham311:
[Notion task](https://www.notion.so/uwblueprintexecs/396370c5e09f4c28bc7f455a7e84c908?v=3751aa62a2594c74b7e0ecb363304c24&p=91ddc3a4b6574688b7d89bc59adf9dcc)

## Changes
- Added in DELETED field for status enum in `schema/patient.js`
- Added in mandatory barcode and triageLevel fields inline with frontend changes
- Added migration to make barcodeValue and triageLevel in the `patients` table to be non-nullable

## Testing
- Added and edited patients using DELETED status
- Added patients without barcode and triageLevel ensuring that GraphQL threw error
- To test the migration:
    1. Enter the database and run the following SQL command: `SELECT column_name, data_type, is_nullable FROM information_schema.columns WHERE TABLE_NAME='patients';`. This tells you which columns are nullable and non-nullable. Pay attention to `barcodeValue` and `triageLevel`.
    2. Run the migration from the shell
    3. Enter the database and re-run the above SQL command. Notice how the nullability has changed for `barcodeValue` and `triageLevel`. You can retry these steps when you migrate down

### DIFF
--- a/migrations/20200707221810-mandatory-patient-fields.js
+++ b/migrations/20200707221810-mandatory-patient-fields.js
@@ -1,0 +1,51 @@
+"use strict";
+
+module.exports = {
+  up: (queryInterface, Sequelize) => {
+    return queryInterface.sequelize.transaction((t) => {
+      return Promise.all([
+        queryInterface.changeColumn("patients", "barcodeValue", {
+          type: Sequelize.BIGINT,
+          allowNull: false,
+        }),
+        queryInterface.sequelize.query(
+          'ALTER TABLE patients ALTER COLUMN "triageLevel" TYPE "enum_patients_triageLevel", ALTER COLUMN "triageLevel" SET NOT NULL'
+        ),
+        // queryInterface.changeColumn("patients", "triageLevel", {
+        //   type: Sequelize.DataTypes.ENUM(
+        //     "WHITE",
+        //     "GREEN",
+        //     "YELLOW",
+        //     "RED",
+        //     "BLACK"
+        //   ),
+        //   allowNull: false,
+        // }),
+      ]);
+    });
+  },
+
+  down: (queryInterface, Sequelize) => {
+    return queryInterface.sequelize.transaction((t) => {
+      return Promise.all([
+        queryInterface.changeColumn("patients", "barcodeValue", {
+          type: Sequelize.BIGINT,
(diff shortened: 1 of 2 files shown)


migrations/20200707221810-mandatory-patient-fields.js
```diff
@@ -0,0 +1,51 @@
...
+
+module.exports = {
+  up: (queryInterface, Sequelize) => {
+    return queryInterface.sequelize.transaction((t) => {
+      return Promise.all([
+        queryInterface.changeColumn("patients", "barcodeValue", {
+          type: Sequelize.BIGINT,
+          allowNull: false,
+        }),
+        queryInterface.sequelize.query(
```
petorsong, line 11: you're right that this is the best way to do this, i truly hate sequelize enums with a passion now

migrations/20200707221810-mandatory-patient-fields.js
```diff
@@ -0,0 +1,51 @@
...
+          'ALTER TABLE patients ALTER COLUMN "triageLevel" TYPE "enum_patients_triageLevel", ALTER COLUMN "triageLevel" SET NOT NULL'
+        ),
+        // queryInterface.changeColumn("patients", "triageLevel", {
```
petorsong, line 14: we can scrap the comments now

migrations/20200707221810-mandatory-patient-fields.js
```diff
@@ -0,0 +1,51 @@
+"use strict";
+
+module.exports = {
+  up: (queryInterface, Sequelize) => {
+    return queryInterface.sequelize.transaction((t) => {
+      return Promise.all([
+        queryInterface.changeColumn("patients", "barcodeValue", {
+          type: Sequelize.BIGINT,
+          allowNull: false,
+        }),
```
petorsong, line 10: turns out the correct way to use enums (probably) is appending `{ transaction: t }` as an argument to the `queryInterface` call, so make sure to do that for all 4 calls you have
```suggestion
        queryInterface.changeColumn("patients", "barcodeValue", {
          type: Sequelize.BIGINT,
          allowNull: false,
        }, { transaction: t }),
```

PR #48, opened 2020-07-17
Location Pins

aaronabraham311:
[Notion task](https://www.notion.so/uwblueprintexecs/396370c5e09f4c28bc7f455a7e84c908?v=3751aa62a2594c74b7e0ecb363304c24&p=09b772f158924f9c8d4196a723e28a9a)

## Changes
- Added table `locationPins` to store location-based information for pins added on maps
- Added resolvers (add, update, delete) and querying functionality
- Added seeders

## Testing
- Steps:
    - Run migration (should already be migrated if using new builder)
    - Run seeder
    - Go ham on GraphIQL

## Screenshots
Multi-pin querying
Screen Shot 2020-07-16 at 7 43 10 PM

Single-pin querying
Screen Shot 2020-07-16 at 7 43 33 PM

Querying pins belonging to a particular event
Screen Shot 2020-07-18 at 11 04 55 AM

Adding pin
Screen Shot 2020-07-16 at 7 45 55 PM

Adding pin error - no associated event
Screen Shot 2020-07-16 at 7 46 17 PM

Adding pin error - no latitude given (same error applies for longitude)
Screen Shot 2020-07-16 at 7 46 46 PM

Updating pin
Screen Shot 2020-07-16 at 7 47 34 PM

Updating pin error - no given pin ID
 Screen Shot 2020-07-16 at 7 47 07 PM

Deleting pin
Screen Shot 2020-07-16 at 7 48 24 PM

Deleting non-existant pin
Screen Shot 2020-07-16 at 7 48 14 PM

### DIFF
--- a/migrations/20200715014355-create-location-pins.js
+++ b/migrations/20200715014355-create-location-pins.js
@@ -32,6 +32,10 @@ module.exports = {
         allowNull: false,
         type: Sequelize.DATE,
       },
+      deletedAt: {
+        allowNull: true,
+        type: Sequelize.DATE,
+      },
     });
   },
   down: (queryInterface, Sequelize) => {
--- a/migrations/20200715014355-create-location-pins.js
+++ b/migrations/20200715014355-create-location-pins.js
@@ -1,0 +1,40 @@
+"use strict";
+
+module.exports = {
+  up: (queryInterface, Sequelize) => {
+    return queryInterface.createTable("locationPins", {
+      id: {
+        allowNull: false,
+        autoIncrement: true,
+        primaryKey: true,
+        type: Sequelize.INTEGER,
+      },
+      label: {
+        type: Sequelize.STRING,
+      },
+      latitude: {
+        type: Sequelize.DOUBLE,
+      },
+      longitude: {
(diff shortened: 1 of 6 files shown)


resolvers/locationPin.js
```diff
@@ -0,0 +1,75 @@
+"use strict";
+
+const db = require("../models");
+
+const locationPinResolvers = {
+  Query: {
```
petorsong, line 6: can we add another query that fetches all pins for a given eventid (something like `pinsForEvent(eventId)`) - in general we will only care about pins for a specific event and care less about getting all pins for all events

migrations/20200715014355-create-location-pins.js
```diff
@@ -0,0 +1,40 @@
+"use strict";
+
+module.exports = {
+  up: (queryInterface, Sequelize) => {
+    return queryInterface.createTable("locationPins", {
```
petorsong, line 5: since this is a new table we're adding, i'd like to try out paranoid tables - it basically implements soft deletes which we want but aren't doing with any existing table currently
https://sequelize.org/master/manual/paranoid.html

PR #49, opened 2020-07-19
users name schema

carelynntsai:
[Notion task](https://www.notion.so/uwblueprintexecs/users-name-schema-d08f5466736d4ac48d2dccbcc1e9e404)

## Changes
- removed "firstName" and "lastName" fields from user type and created a single field "name"
- updated seeders 
- updates user schema, model, resolvers
- created new migration

## Testing
- seeded data
- tested migration up and down functions
- tested resolvers

## Screenshots
image
image
image

### DIFF
--- a/migrations/20200719025533-add-name-column-users.js
+++ b/migrations/20200719025533-add-name-column-users.js
@@ -12,9 +12,13 @@ module.exports = {
         { transaction: t }
       )
       await queryInterface.sequelize.query(
-        `UPDATE users SET "name" = CONCAT("firstName", ' ', "lastName")`,
+        `UPDATE users SET "name" = "firstName" WHERE "lastName"=''`,
         { transaction: t }
-      )
+      );
+      await queryInterface.sequelize.query(
+        `UPDATE users SET "name" = CONCAT("firstName", ' ', "lastName") WHERE "lastName"<>''`,
+        { transaction: t }
+      );
       await queryInterface.removeColumn('users', 'firstName', { transaction: t });
       return queryInterface.removeColumn('users', 'lastName', { transaction: t });
     });
--- a/migrations/20200719025533-add-name-column-users.js
+++ b/migrations/20200719025533-add-name-column-users.js
@@ -1,0 +1,48 @@
+'use strict';
+
+module.exports = {
+  up: (queryInterface, Sequelize) => {
+    return queryInterface.sequelize.transaction(async (t) => {
+      await queryInterface.addColumn(
+        "users",
+        "name",
+        {
+          type: Sequelize.DataTypes.STRING,
+        },
+        { transaction: t }
+      )
+      await queryInterface.sequelize.query(
+        `UPDATE users SET "name" = CONCAT("firstName", ' ', "lastName")`,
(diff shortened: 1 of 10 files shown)


carelynntsai:
@petorsong wasn't sure if I should set the column to allowNull: false, the previous firstName/lastName columns weren't

migrations/20200719025533-add-name-column-users.js
```diff
@@ -0,0 +1,48 @@
...
+        "name",
+        {
+          type: Sequelize.DataTypes.STRING,
+        },
+        { transaction: t }
+      )
+      await queryInterface.sequelize.query(
+        `UPDATE users SET "name" = CONCAT("firstName", ' ', "lastName")`,
+        { transaction: t }
+      )
```
petorsong, line 17: i added a [migration that adds an admin user](https://github.com/uwblueprint/paramedics-web/blob/master/migrations/20200717012421-load-admin-user.js) before this change, which sets `lastName` to an empty string, which appends a space once this migration is run - this suggestion is to account for that edge case, but test it first cause i dunno if it's quite accurate

```suggestion
      await queryInterface.sequelize.query(
        `UPDATE users SET "name" = "firstName" WHERE "lastName"=''`,
        { transaction: t }
      );
      await queryInterface.sequelize.query(
        `UPDATE users SET "name" = CONCAT("firstName", ' ', "lastName") WHERE "lastName"<>''`,
        { transaction: t }
      );
```

migrations/20200719025533-add-name-column-users.js
```diff
@@ -0,0 +1,48 @@
+'use strict';
+
+module.exports = {
+  up: (queryInterface, Sequelize) => {
+    return queryInterface.sequelize.transaction(async (t) => {
```
petorsong, line 5: is there a particular need to use this await/async syntax instead of the `Promise.all` syntax?
https://sequelize.org/master/manual/migrations.html#migration-skeleton

migrations/20200719025533-add-name-column-users.js
```diff
@@ -0,0 +1,48 @@
...
+      await queryInterface.addColumn(
+        "users",
+        "lastName",
+        {
+          type: Sequelize.DataTypes.STRING,
+        },
+        { transaction: t }
+      );
+      await queryInterface.sequelize.query(
+        'UPDATE users SET "firstName" = "name", "lastName" = "name"',
```
petorsong, line 42: this isn't super sustainable - the names get too long if you run undo a bunch of times (like i did during testing); if you can find an easy way to do a slightly different operation (e.g. explode `name` on ' ' and put the first part in `firstName` and second part in `lastName`) that would be preferred, otherwise just lemme know and we'll flag it to be fixed later

carelynntsai: would it be better to do a similar thing like you did with admin and ``UPDATE users SET "firstName" = "name", "lastName" = ''``

petorsong: let's just leave it as is, but maybe add a comment here to say running this down migration too often may result in issues

PR #51, opened 2020-07-19
change barcodeValue to string and add corresponding migrations

JasonYG:
[Notion task](https://www.notion.so/uwblueprintexecs/396370c5e09f4c28bc7f455a7e84c908?v=3751aa62a2594c74b7e0ecb363304c24&p=0df46e3a5d1a40a5a9c685852a9ba0e3)

## Changes
- Changed schema and model of `barcodeValue` from `int` to `string`
- Added migration to remove the `barcodeValue` column
- Added second migration to add the `barcodeValue` column as a string

## Testing
- Manual testing on graphql playground
- used `addPatient` mutation to create patients with a string `barcodeValue`


migrations/20200719172434-remove-patients-barcodeValue.js
```diff
@@ -0,0 +1,24 @@
+"use strict";
+
+module.exports = {
+  up: (queryInterface, Sequelize) => {
+    return queryInterface.sequelize.transaction((t) => {
+      return Promise.all([
```
petorsong, line 6: these two migrations should be in the same file

migrations/20200719172434-change-patients-barcodeValue-to-string.js
```diff
@@ -0,0 +1,39 @@
+"use strict";
+
+module.exports = {
+  up: (queryInterface, Sequelize) => {
```
petorsong, line 4: as it stands, this migration will wipe any existing values in the `barcodeValue` column - we don't want this to happen, so ideally you migrate those int values to string so we can keep the values still

migrations/20200719172434-change-patients-barcodeValue-to-string.js
```diff
@@ -0,0 +1,39 @@
...
+          {
+            type: Sequelize.DataTypes.STRING,
+          },
+          { transaction: t }
+        ),
+      ]);
+    });
+  },
+
+  down: (queryInterface, Sequelize) => {
```
petorsong, line 22: similar issue here, except converting from a string to an int is not a clean operation - to keep things simple, i would only convert the string values that consist of only numbers (i.e. all "string" integers)

if you're having trouble with this, just ping me on slack

JasonYG: fixed in latest commit, but not sure if it's correct

migrations/20200719172434-change-patients-barcodeValue-to-string.js
```diff
@@ -0,0 +1,33 @@
...
+    return queryInterface.sequelize.transaction((t) => {
+      return Promise.all([
+        queryInterface.changeColumn(
+          "patients",
+          "barcodeValue",
+          { type: 'INTEGER USING CAST("barcodeValue" as INTEGER)' },
```
JasonYG, line 25: do you think this is the right way of doing this? I found on stack overflow lol

migrations/20200719172434-change-patients-barcodeValue-to-string.js
```diff
@@ -0,0 +1,33 @@
...
+          {
+            transaction: t,
+          }
+        ),
+      ]);
+    });
+  },
```
petorsong, line 32: this will set any barcodes that aren't integers to `null` which will ensure the `CAST` will always work; this isn't perfect but considering this down migration shouldn't be run very often, it should be an ok middle ground

**please test this out yourself before committing**
```suggestion
    up: (queryInterface, Sequelize) => {
    return queryInterface.changeColumn(
      "patients", "barcodeValue", { type: Sequelize.DataTypes.STRING }
    );
  },

  down: (queryInterface, Sequelize) => {
    return queryInterface.sequelize.transaction((t) => {
      return Promise.all([
        queryInterface.bulkUpdate('patients',
          { barcodeValue: null },
          { barcodeValue: { [Sequelize.Op.regexp]: '[^0-9]' } },
          { transaction: t }
        ),
        queryInterface.changeColumn(
          "patients",
          "barcodeValue",
          { type: 'INTEGER USING CAST("barcodeValue" as INTEGER)' },
          {
            transaction: t,
          }
        ),
      ]);
    });
  },
```

JasonYG: yup tested it, it works! Thanks

PR #53, opened 2020-07-26
Filter Queries

aaronabraham311:
[Notion task](https://www.notion.so/uwblueprintexecs/396370c5e09f4c28bc7f455a7e84c908?v=3751aa62a2594c74b7e0ecb363304c24&p=085793a76ccd4f5a9dec0326d2bdfe6e)

## Changes
- Created query to filter collection points by event ID
- Created query to filter patients by collection points

## Testing
- Tested through GraphIQL
- Checked DB 

## Screenshots
Screen Shot 2020-07-26 at 1 41 38 PM
Screen Shot 2020-07-26 at 1 42 48 PM

### DIFF
--- a/resolvers/ambulance.js
+++ b/resolvers/ambulance.js
@@ -9,15 +9,6 @@ const ambulanceResolvers = {
         include: [
           {
             model: db.event,
-            attributes: [
-              "id",
-              "name",
-              "eventDate",
-              "createdBy",
-              "isActive",
-              "createdAt",
-              "updatedAt",
-            ],
           },
         ],
       }),
@@ -26,15 +17,6 @@ const ambulanceResolvers = {
         include: [
           {
             model: db.event,
-            attributes: [
-              "id",
-              "name",
-              "eventDate",
-              "createdBy",
-              "isActive",
-              "createdAt",
-              "updatedAt",
-            ],
           },
         ],
       }),
--- a/resolvers/ambulance.js
+++ b/resolvers/ambulance.js
@@ -4,29 +4,65 @@ const db = require("../models");
 
 const ambulanceResolvers = {
   Query: {
-    ambulances: () => db.ambulance.findAll(),
-    ambulance: (obj, args, context, info) => db.ambulance.findByPk(args.id)
+    ambulances: () =>
+      db.ambulance.findAll({
+        include: [
+          {
+            model: db.event,
+            attributes: [
+              "id",
+              "name",
+              "eventDate",
+              "createdBy",
+              "isActive",
+              "createdAt",
+              "updatedAt",
+            ],
+          },
+        ],
+      }),
+    ambulance: (obj, args, context, info) =>
+      db.ambulance.findByPk(args.id, {
+        include: [
+          {
+            model: db.event,
+            attributes: [
+              "id",
+              "name",
+              "eventDate",
+              "createdBy",
+              "isActive",
+              "createdAt",
+              "updatedAt",
+            ],
+          },
+        ],
+      }),
   },
   Mutation: {
     addAmbulance: (parent, args) => {
-        return db.ambulance.create({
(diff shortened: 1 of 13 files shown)


resolvers/collectionPoint.js
```diff
@@ -1,87 +1,89 @@
...
-
-    collectionPoint: {
-        eventId: (obj, args, context, info) => db.event.findByPk(obj.eventId),
-        createdBy: (obj, args, context, info) => db.user.findByPk(obj.createdBy)
-    },
+  Query: {
+    collectionPoints: () => db.collectionPoint.findAll(),
+    collectionPoint: (obj, args, context, info) =>
+      db.collectionPoint.findByPk(args.id),
+    collectionPointsByEventId: (obj, args, context, info) =>
```
petorsong, line 10: nitpick: `collectionPointsByEvent` is a better format, no need to refer to the id in the query name
```suggestion
    collectionPointsByEvent: (obj, args, context, info) =>
```

resolvers/patient.js
```diff
@@ -1,88 +1,99 @@
...
-        },
+  Query: {
+    patients: () => db.patient.findAll(),
+    patient(obj, args, context, info) {
+      return db.patient.findByPk(args.id);
     },
-    Patient: {
-        collectionPointId: (obj, args, context, info) => db.collectionPoint.findByPk(obj.collectionPointId),
-        hospitalId: (obj, args, context, info) => db.hospital.findByPk(obj.hospitalId)
+    patientsByCCPId: (obj, args, context, info) =>
```
petorsong, line 11: same deal, also i'm a fan of true pascal case
```suggestion
    patientsByCcp: (obj, args, context, info) =>
```

schema/collectionPoint.js
```diff
@@ -4,6 +4,7 @@ extend type Query {
 
     collectionPoint(id: ID!): collectionPoint
     collectionPoints: [collectionPoint]
+    collectionPointsByEventId(eventId: ID!): [collectionPoint]
```
petorsong, line 7:
```suggestion
    collectionPointsByEvent(eventId: ID!): [collectionPoint]
```

schema/patient.js
```diff
@@ -17,6 +17,7 @@ const patientSchema = `
   extend type Query {
     patient(id: ID!): Patient
     patients: [Patient]
+    patientsByCCPId(collectionPointId: ID!): [Patient]
```
petorsong, line 20: do we have a pattern of calling the id `collectionPointId` already (like with the full entity name)? if not i would prefer `ccpId`
```suggestion
    patientsByCcp(collectionPointId: ID!): [Patient]
```

aaronabraham311: We do, check line 29 and line 44

resolvers/ambulance.js
```diff
@@ -4,47 +4,83 @@ const db = require("../models");
...
 const ambulanceResolvers = {
   Query: {
-    ambulances: () => db.ambulance.findAll(),
-    ambulance: (obj, args, context, info) => db.ambulance.findByPk(args.id)
+    ambulances: () =>
+      db.ambulance.findAll({
+        include: [
+          {
+            model: db.event,
+            attributes: [
```
petorsong, line 12: is there a reason we need to list all the attributes here (this is done a couple of times)? this example i found seems to only have the model
https://gist.github.com/zcaceres/83b554ee08726a734088d90d455bc566#vanilla-include

aaronabraham311: I had no idea that we could do this. I assume that Sequelize parses this into a `SELECT *`. I removed the attribute list from queries with this similar behaviours.

petorsong:
just wanna double check these changes don't affect https://github.com/uwblueprint/paramedics-web/pull/54 (or vice versa) before merging

PR #55, opened 2020-07-30
Soft Deletion

aaronabraham311:
[Notion task](https://www.notion.so/uwblueprintexecs/396370c5e09f4c28bc7f455a7e84c908?v=3751aa62a2594c74b7e0ecb363304c24&p=fad01026a76541aca00293d1d6b07ac1)

[Paranoid tables documentation](https://www.notion.so/uwblueprintexecs/6a0ad062ce504acc90fa25cb4d2c731f?v=4844a823da724b7ebf2c442a1ee1279f&p=4fabc50d4a9b4c969c96d7ae5bc19cfc)

## Changes
- Converted all tables into paranoid tables
- Added restoration resolvers for all tables 
- Created cascade delete and restoration functionality for:
    - Events -> CCP -> Patients: if event is deleted, all associated CCP and patients are also deleted. If an event is restored, all associated CCP and patients are restored as well. Same applies to CCP -> patients.
    - Events <-> Ambulances, Events <-> Hospitals: The association between events and resources is deleted if either the event or the resource is deleted. The association is restored if both the event and the resource are restored
- Query archived events

## Testing
- <b>Please look at the paranoid tables documentation above to understand the changes I have made</b>
- Manual GraphIQL testing combined with DB checking (please use the DB CLI when checking if associations are deleted)

## Screenshots
Archived event query
Screen Shot 2020-07-29 at 6 49 08 PM

Restoring a previously deleted event
Screen Shot 2020-07-29 at 6 50 26 PM

Deleting event#19 and corresponding CCP deletion
Screen Shot 2020-07-29 at 6 52 55 PM

Restoring event#19 and corresponding CCP restoration
Screen Shot 2020-07-29 at 6 53 35 PM

Deleting ambulance #14 and corresponding event-ambulance association deletion
Screen Shot 2020-07-29 at 6 54 54 PM

Deleting event #22 and corresponding event-ambulance deletion
Screen Shot 2020-07-29 at 6 55 57 PM

Restoring event#22 (note that the associations with ambulance #14 are still deleted because ambulance #14 is not restored)
Screen Shot 2020-07-29 at 6 56 54 PM

Restoring ambulance #14 
Screen Shot 2020-07-29 at 6 57 27 PM


petorsong:
also i'm still seeing that `resolvers/patient.js` has merge conflicts according to github - just double check if that's still an issue

models/collectionPoint.js
```diff
@@ -32,7 +32,6 @@ module.exports = (sequelize, DataTypes) => {
   collectionPoint.associate = function (models) {
     // associations can be defined here
     collectionPoint.hasMany(models.patient, {
-      onDelete: "CASCADE",
```
petorsong: why was this removed? is the `CASCADE` here unnecessary?

models/eventAmbulances.js
```diff
@@ -10,7 +10,8 @@ module.exports = (sequelize, DataTypes) => {
     { paranoid: true }
   );
   eventAmbulances.associate = function (models) {
-    // define associations here
+    eventAmbulances.belongsTo(models.event);
+    eventAmbulances.belongsTo(models.ambulance);
```
petorsong, line 14: i'm guessing this is introduced to optimize the fetch calls - basically doing the `JOIN` stuff i mentioned previously?

resolvers/hospital.js
```diff
@@ -54,29 +54,27 @@ const hospitalResolvers = {
...
       const associatedEvents = await db.eventHospitals.findAll({
         where: {
           hospitalId: args.id,
         },
+        include: [
+          {
+            model: db.event,
+            required: true,
+          },
+        ],
```
petorsong, line 67: do we need to explicitly require the `event` model here? i thought `eventHospitals` would already provide the event ID and hospital ID for free?

models/collectionPoint.js
aaronabraham311: No, functionally it has no impact. I was using this when I was experimenting with cascade deletes on non-paranoid tables.

models/eventAmbulances.js
aaronabraham311: Exactly. It introduces the functionality needed to do the joins.

resolvers/hospital.js
aaronabraham311: No we need to specify it in this manner to instruct Sequelize to perform an inner join. Otherwise, it will just give us the `eventHospitals` without joins.

models/collectionPoint.js
```diff
@@ -1,29 +1,50 @@
...
+          instance.getPatients().then((patients) =>
+            patients.map((patient) => {
+              patient.destroy();
+            })
+          );
+        },
+        afterRestore: function (instance, options) {
+          instance
+            .getPatients({ paranoid: false })
+            .then((patients) => patients.map((patient) => patient.restore()));
```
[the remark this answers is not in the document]
aaronabraham311: Good question. Correct me if I am wrong @petorsong but we had discussed this exact case. On the front-end, there is no functionality to system-delete a patient. Instead, the user will have the option to change the status of a patient to `DELETED`. This means that the only way to system-delete a patient is if we delete a collection point. If we didn't do this, then you would be right: we would have no past knowledge whether we deleted a patient through a cascade or manually deleted a single patient.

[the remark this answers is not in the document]
petorsong: yep the reason we don't want to use the actual soft delete for patients is because we wanted to avoid this ambiguity between deleted patients and patients with the `DELETED` status - we will only be using the latter for now, but also have the former available if we do want to use it later on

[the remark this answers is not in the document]
danielchen115: Ah okay, thanks for the clarification.

PR #60, opened 2020-08-12
Add linting

petorsong:
[Notion task](https://www.notion.so/uwblueprintexecs/docker-linter-setup-e63c6fbc895b4f4e904e228bab13582b)

**SEE MY INLINE COMMENTS FOR MOST RELEVANT CHANGES**

## Changes
- linted all files based on linter set up in this commit: https://github.com/uwblueprint/paramedics-web/commit/6c05f55f53b0c6d9042e1d5ba5f46aea1c5e486f
  - refactored some code for linting and performance
- reverted `eventAmbulances` and `eventHospitals` tables to be non-paranoid
  - refactored `addAmbulancesToEvent` and `addHospitalsToEvent` as they were really unwieldy with soft deletes
- set up `dev-lint` job in circleCI
  - refactored existing jobs/steps for performance
    - `dev-build` builds using dev dockerfile
    - `prod-build` builds using prod dockerfile (new)
    - `dev-lint` lints using dev dockerfile as that's where we installed eslint (new)
    - removed `install` step as it wasn't doing anything

## Testing
- running `npm run lint` locally (or via vscode) has no errors
- verify that existing resolvers are working as expected (especially `event`)
- test out `addAmbulancesToEvent` and `addHospitalsToEvent`
  - add and delete resources and check the db tables to make sure all changes look correct
- look with your eyes at the checks on this PR and see that all jobs have a green check mark

## Notes
- we should move towards preferring promise chaining over `await` statements - the latter can still be used when necessary
- in general, our CI/CD pipeline is really small and thus fairly fast - we can break down the steps in our `docker-compose` file and optimize even more for performance, but it's not that expensive (i.e. slow) to just do everything from scratch with each build
- not entirely convinced that the caching within jobs are even doing anything - removed for now
  - this is related to the fact that we simply use `docker-compose up` to build the container, which would not get the benefit of caching `package-lock.json` - we could optimize by breaking it down into discrete steps ([more info here](https://circleci.com/docs/2.0/caching/#section=projects))

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -56,7 +56,7 @@ jobs:
       - restore_cache:
           keys:
             - dev-dependencies-{{ checksum "package.json" }}
-              - dev-dependencies-
+            - dev-dependencies-
       - setup_remote_docker:
           docker_layer_caching: false
       - node/with-cache:
@@ -105,8 +105,22 @@ jobs:
           keys:
             - dev-dependencies-{{ checksum "package.json" }}
               - dev-dependencies-
-      - attach_workspace:
-          at: .
+      - setup_remote_docker:
+          docker_layer_caching: false
+      - run:
+          name: Install docker-compose
+          command: |
+            set -x
+            curl -L https://github.com/docker/compose/releases/download/1.12.0/docker-compose-`uname -s`-`uname -m` > /usr/local/bin/docker-compose
+            chmod +x /usr/local/bin/docker-compose
+      - run:
+          name: Add mock env file
+          command: touch paramedics-web.env
+      - run: 
+          name: Start container via docker-compose
+          command: |
+            set -x
+            docker-compose -f docker-compose-dev.yaml up -d
       - run:
           name: Lint files
           command: npm run lint
@@ -125,4 +139,4 @@ workflows:
             - dev-install
       - dev-lint:
           requires:
-            - dev-build
+            - dev-install
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -53,7 +53,4 @@ workflows:
   version: 2
   build-test:
     jobs:
-      - install
-      - build:
-          requires:
-            - install
+      - build
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -50,26 +50,6 @@ jobs:
           paths:
             - node_modules
           key: prod-dependencies-{{ checksum "package.json" }}
-          
-  prod-install:
-    docker:
-      - image: uwbpparamedics/uwblueprint-paramedics-web:0.0.1
-    steps:
-      - checkout
-      - restore_cache:
-          keys:
-            - prod-dependencies-{{ checksum "package.json" }}
-              - prod-dependencies-
-      - setup_remote_docker:
-          docker_layer_caching: false
-      - node/with-cache:
-          steps:
-            - run: npm install
-      - save_cache:
-          paths:
-            - node_modules
-          key: prod-dependencies-{{ checksum "package.json" }}
-
 
   
 workflows:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -90,6 +90,9 @@ jobs:
       - run:
           name: Building image
           command: docker build -t uwbpparamedics/uwblueprint-paramedics-web:0.0.1 -f Dockerfile.dev .
+      - run:
+          name: Add mock env file
+          command: touch paramedics-web.env
       - run: 
           name: Start container via docker-compose
           command: |
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -104,6 +104,4 @@ workflows:
   dev-build:
     jobs:
       - dev-build
-      - dev-lint:
-          requires:
-            - dev-build
+      - dev-lint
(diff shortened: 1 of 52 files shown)


migrations/20200812182030-non-paranoid-resource-associations.js
```diff
@@ -0,0 +1,41 @@
+'use strict';
```
petorsong, line 1: new migration to make junction tables non-paranoid

models/collectionPoint.js
```diff
@@ -1,48 +1,42 @@
...
-    // associations can be defined here
+  collectionPoint.associate = (models) => {
     collectionPoint.hasMany(models.patient, {
       hooks: true,
     });
 
-    collectionPoint.belongsTo(User(sequelize, DataTypes), {
-      foreignKey: "createdBy",
-      targetKey: "id",
+    collectionPoint.belongsTo(models.user, {
```
petorsong, line 32: refactored so that the function arg `models` was being used and we wouldn't need to import `User` separately

models/eventAmbulances.js
```diff
@@ -1,15 +1,11 @@
...
+'use strict';
 
 module.exports = (sequelize, DataTypes) => {
-  const eventAmbulances = sequelize.define(
-    "eventAmbulances",
-    {
-      eventId: DataTypes.INTEGER,
-      ambulanceId: DataTypes.INTEGER,
-    },
-    { paranoid: true }
```
petorsong, line 10: `paranoid` gone

models/eventHospitals.js
```diff
@@ -1,15 +1,11 @@
...
+'use strict';
 
 module.exports = (sequelize, DataTypes) => {
-  const eventHospitals = sequelize.define(
-    "eventHospitals",
-    {
-      eventId: DataTypes.INTEGER,
-      hospitalId: DataTypes.INTEGER,
-    },
-    { paranoid: true }
```
petorsong, line 10: `paranoid` gone

models/patient.js
```diff
@@ -1,25 +1,21 @@
...
+    'patient',
     {
       gender: DataTypes.STRING,
       age: DataTypes.INTEGER,
       runNumber: DataTypes.BIGINT,
       barcodeValue: DataTypes.STRING,
       status: {
         type: DataTypes.ENUM,
-        values: ["ON_SITE", "RELEASED", "TRANSPORTED"],
+        values: ['ON_SITE', 'RELEASED', 'TRANSPORTED', 'DELETED'],
```
petorsong, line 13: missing for some reason but added back in

resolvers/event.js
```diff
@@ -1,6 +1,6 @@
-"use strict";
+'use strict';
```
petorsong, line 1: this is the main file with most of the refactoring

resolvers/event.js
```diff
@@ -34,15 +34,16 @@ const eventResolvers = {
...
   Mutation: {
     addEvent: async (parent, args) => {
       // Check if createdBy is valid
-      const user = await db.user.findByPk(args.createdBy);
-      if (!user) {
-        throw new Error("Invalid user ID");
-      }
+      await db.user.findByPk(args.createdBy).then((user) => {
+        if (!user) {
+          throw new Error('Invalid user ID: ' + args.createdBy);
```
petorsong, line 44: added the invalid value to error messages for this file - [we have a ticket to do this for all other error messages](https://www.notion.so/uwblueprintexecs/resolver-cleanup-ad3e0ae00db24983ac0c0722dd6c6400)

resolvers/event.js
```diff
@@ -62,27 +63,33 @@ const eventResolvers = {
...
-        // Checking if ambulances are valid
-        for (const ambulanceId of args.ambulances) {
-          const ambulance = await db.ambulance.findByPk(ambulanceId["id"]);
-          if (!ambulance) {
-            return true;
-          }
-        }
+      if (args.ambulances) {
+        // Checking if all ambulances exist
+        await Promise.all(
```
petorsong, line 83: this syntax makes sure all promises (i.e. `db` calls) returned from the map resolve first (via await) before proceeding

resolvers/event.js
```diff
@@ -220,47 +213,47 @@ const eventResolvers = {
...
-          await db.eventHospitals.create({
-            eventId: args.eventId,
-            hospitalId: hospitalId["id"],
-          });
-        } else {
-          await db.eventHospitals.restore({
-            where: {
-              eventId: args.eventId,
-              hospitalId: hospitalId["id"],
-            },
```
petorsong, line 258: no need to restore anymore if we don't have paranoid associations

server.js
```diff
@@ -6,5 +6,6 @@ const { schema } = require('./graphql');
 const server = new ApolloServer({ schema });
 
 server.listen().then(({ url }) => {
+  // eslint-disable-next-line no-console
```
petorsong, line 9: mildly annoying but we'll keep this around for now

.circleci/config.yml
```diff
@@ -37,23 +35,31 @@ jobs:
...
+  build-and-lint:
     jobs:
-      - install
-      - build:
-          requires:
-            - install
\ No newline at end of file
+      - prod-build
+      - dev-build
+      - dev-lint
```
petorsong: note that all three jobs are independent - there wasn't any reason to keep them linked together

aaronabraham311: Hmm surprised that you can do `dev-lint` in parallel to `dev-build`

petorsong: How this works is still kinda murky to me but at this point I'm not focused on getting the best config possible - just one that does what we want relatively quickly (I'm sure there's further ways to optimize this but it's not worth the effort right now)

PR #62, opened 2020-08-24
Gender changes for patient

aaronabraham311:
[Notion task](https://www.notion.so/uwblueprintexecs/4b97736541b2418eac9025b0efcf3c58?v=f07847298f194931a3f9052e205544e5&p=fd09d01ba3984ae6a9d07b9fa71c363b)

## Changes
- Created migration that:
   - Converts all prexisting "Male" and "Female" genders to "M" and "F" among patients
   - Created an enum for gender with values of "M" and "F"
- Updated GraphQL schema to reflect gender enum
- Updated patient seeders to remove gender fields that were invalid (i.e. "unknown"). 

## Testing
- Since there are seeding changes in this PR, make sure you undo all seeders and re-seed the database
- Migrated and un-migrated the enum several times
    - For each of the migrations, I made sure that the database updated accordingly
- Forced errors by providing wrong enum values on GraphIQL.
## Screenshots
Screen Shot 2020-08-26 at 7 51 34 AM

Screen Shot 2020-08-26 at 7 52 03 AM

### DIFF
--- a/migrations/20200824194331-patient-gender.js
+++ b/migrations/20200824194331-patient-gender.js
@@ -11,7 +11,8 @@ module.exports = {
             gender: {
               [Sequelize.Op.notIn]: ['F', 'Female', 'Male', 'M'],
             },
-          }
+          },
+          { transaction: t }
         ),
       ])
         .then(
--- a/migrations/20200824194331-patient-gender.js
+++ b/migrations/20200824194331-patient-gender.js
@@ -14,33 +14,30 @@ module.exports = {
           },
           { transaction: t }
         ),
-      ])
-        .then(
-          queryInterface.bulkUpdate(
-            'patients',
-            { gender: 'F' },
-            { gender: 'Female' },
-            { transaction: t }
-          ),
-          queryInterface.bulkUpdate(
-            'patients',
-            { gender: 'M' },
-            { gender: 'Male' },
-            { transacton: t }
-          )
+        queryInterface.bulkUpdate(
+          'patients',
+          { gender: 'F' },
+          { gender: 'Female' },
+          { transaction: t }
+        ),
+        queryInterface.bulkUpdate(
+          'patients',
+          { gender: 'M' },
+          { gender: 'Male' },
+          { transacton: t }
+        ),
+      ]).then(() =>
+        queryInterface.changeColumn(
+          'patients',
+          'gender',
+          {
+            type: Sequelize.DataTypes.ENUM('M', 'F'),
+          },
+          {
+            transaction: t,
+          }
         )
-        .then(() =>
-          queryInterface.changeColumn(
-            'patients',
-            'gender',
-            {
-              type: Sequelize.DataTypes.ENUM('M', 'F'),
-            },
-            {
-              transaction: t,
-            }
-          )
-        );
+      );
     });
   },
 
--- a/migrations/20200824194331-patient-gender.js
+++ b/migrations/20200824194331-patient-gender.js
@@ -7,21 +7,39 @@ module.exports = {
         queryInterface.bulkUpdate(
           'patients',
           { gender: 'F' },
-          { gender: { [Sequelize.Op.regexp]: '.*' } },
-          { transaction: t }
-        ),
-      ]).then(() =>
-        queryInterface.changeColumn(
-          'patients',
-          'gender',
           {
-            type: Sequelize.DataTypes.ENUM('M', 'F'),
-          },
-          {
-            transaction: t,
+            gender: {
+              [Sequelize.Op.notIn]: ['F', 'Female', 'Male', 'M'],
+            },
           }
-        )
-      );
+        ),
+      ])
+        .then(() => {
+          queryInterface.bulkUpdate(
+            'patients',
+            { gender: 'F' },
+            { gender: 'Female' },
+            { transaction: t }
+          ),
+            queryInterface.bulkUpdate(
+              'patients',
+              { gender: 'M' },
+              { gender: 'Male' },
+              { transacton: t }
+            );
+        })
+        .then(() =>
+          queryInterface.changeColumn(
+            'patients',
+            'gender',
+            {
+              type: Sequelize.DataTypes.ENUM('M', 'F'),
+            },
+            {
+              transaction: t,
+            }
+          )
+        );
     });
   },
 
--- a/migrations/20200824194331-patient-gender.js
+++ b/migrations/20200824194331-patient-gender.js
@@ -16,6 +16,7 @@ module.exports = {
           { gender: 'Female' },
           { transaction: t }
         ),
+      ]).then(() =>
         queryInterface.changeColumn(
           'patients',
           'gender',
@@ -25,8 +26,8 @@ module.exports = {
           {
             transaction: t,
           }
-        ),
-      ]);
+        )
+      );
     });
   },
 
--- a/migrations/20200824194331-patient-gender.js
+++ b/migrations/20200824194331-patient-gender.js
@@ -14,20 +14,20 @@ module.exports = {
           }
         ),
       ])
-        .then(() => {
+        .then(
           queryInterface.bulkUpdate(
             'patients',
             { gender: 'F' },
             { gender: 'Female' },
             { transaction: t }
           ),
-            queryInterface.bulkUpdate(
-              'patients',
-              { gender: 'M' },
-              { gender: 'Male' },
-              { transacton: t }
-            );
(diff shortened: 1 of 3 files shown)


schema/patient.js
```diff
@@ -14,6 +14,11 @@ const patientSchema = `
     DELETED
   }
 
+  enum gender {
```
petorsong, line 17: since `Gender` is a type/enum, we should capitalize it as a convention (may also affect frontend schema types)
```suggestion
  enum Gender {
```

schema/patient.js
```diff
@@ -22,7 +27,7 @@ const patientSchema = `
 
   extend type Mutation {
     addPatient(
-      gender: String,
+      gender: gender,
```
petorsong, line 30:
```suggestion
      gender: Gender,
```

schema/patient.js
```diff
@@ -38,7 +43,7 @@ const patientSchema = `
     
     updatePatient(
       id: ID!,
-      gender: String,
+      gender: gender,
```
petorsong, line 46:
```suggestion
      gender: Gender,
```

schema/patient.js
```diff
@@ -59,7 +64,7 @@ const patientSchema = `
   
   type Patient {
     id: ID!,
-    gender: String,
+    gender: gender,
```
petorsong, line 67:
```suggestion
    gender: Gender,
```

seeders/20200206005427-demo-patient.js
```diff
@@ -56,7 +56,7 @@ module.exports = {
         updatedAt: new Date(),
       },
       {
-        gender: 'unknown',
+        gender: 'Male',
```
petorsong, line 59: shouldn't we be using the new enum values now? (same goes for all other `gender` values in seeders)
```suggestion
        gender: 'M',
```

migrations/20200824194331-patient-gender.js
```diff
@@ -0,0 +1,64 @@
...
+          {
+            type: Sequelize.DataTypes.ENUM('M', 'F'),
+          },
+          {
+            transaction: t,
+          }
+        ),
+      ]);
+    });
+  },
```
petorsong, line 31: two issues here:
(1) we don't handle values that aren't `Male` or `Female` here - this means that when we try to do the `changeColumn` operation, any invalid values are still in the table and causes it to error out; for simplicity's sake, let's just take all these values and convert them into `F`

(2) we need the updates to complete first before attempting to changing the column type, otherwise it'll error out if there are any invalid values due to race conditions - transactions don't ensure the order of operations is preserved

the following suggestion only attempts to address (2) and should be tested as well
```suggestion
  up: (queryInterface, Sequelize) => {
    return queryInterface.sequelize.transaction((t) => {
      return Promise.all([
        queryInterface.bulkUpdate(
          'patients',
          { gender: 'M' },
          { gender: 'Male' },
          { transaction: t }
        ),
        queryInterface.bulkUpdate(
          'patients',
          { gender: 'F' },
          { gender: 'Female' },
          { transaction: t }
        ),
      ]).then(() =>
        queryInterface.changeColumn(
          'patients',
          'gender',
          {
            type: Sequelize.DataTypes.ENUM('M', 'F'),
          },
          {
            transaction: t,
          }
        )
      );
    });
  },
```

migrations/20200824194331-patient-gender.js
```diff
@@ -0,0 +1,64 @@
...
+
+  down: (queryInterface, Sequelize) => {
```
petorsong, line 33: same issues from the up migration need to be addressed here - note that non-male/female values can't be recovered so you don't have to worry about that here

migrations/20200824194331-patient-gender.js
aaronabraham311: For condition 1, is this the expected behaviour you wanted?
Screen Shot 2020-08-26 at 1 06 49 PM

petorsong: @aaronabraham311 not quite - it shouldn't end up with all three as `F`; end results should be something like:
`Male` -> `M`
`Female` -> `M`
Other -> `F`

migrations/20200824194331-patient-gender.js
```diff
@@ -0,0 +1,63 @@
...
+module.exports = {
+  up: (queryInterface, Sequelize) => {
+    return queryInterface.sequelize.transaction((t) => {
+      return Promise.all([
+        queryInterface.bulkUpdate(
+          'patients',
+          { gender: 'F' },
+          { gender: { [Sequelize.Op.regexp]: '.*' } },
+          { transaction: t }
+        ),
```
wongcmichelle, line 12: I think this is what we want, right?
```suggestion
        queryInterface.bulkUpdate(
          'patients',
          { gender: 'M' },
          { gender: 'Male' },
          { transaction: t }
        ),
        queryInterface.bulkUpdate(
          'patients',
          { gender: 'F' },
          { gender: 'Female' },
          { transaction: t }
        ),
        queryInterface.bulkUpdate(
          'patients',
          { gender: 'F' },
          { gender: { [Sequelize.Op.regexp]: '.*' } },
          { transaction: t }
        ),
```

petorsong: not quite - i believe the last `bulkUpdate` will just overwrite the first 2, so ideally we be more specific and query for all values that are `NOT 'M' OR 'F'`

additionally, you could do it in this order
- both `Male` -> `M` and `Female` -> `F` first
- then `NOT 'M' OR 'F'`-> `F`
- then change column to use new enum

migrations/20200824194331-patient-gender.js
```diff
@@ -7,21 +7,39 @@ module.exports = {
...
-          'patients',
-          'gender',
-          {
-            type: Sequelize.DataTypes.ENUM('M', 'F'),
-          },
           {
-            transaction: t,
+            gender: {
+              [Sequelize.Op.notIn]: ['F', 'Female', 'Male', 'M'],
+            },
```
wongcmichelle, line 13: did you intentionally remove `transaction: t,` here?

aaronabraham311: Good catch, should be fixed now

migrations/20200824194331-patient-gender.js
```diff
@@ -0,0 +1,82 @@
...
+            'patients',
+            'gender',
+            {
+              type: Sequelize.DataTypes.ENUM('M', 'F'),
+            },
+            {
+              transaction: t,
+            }
+          )
+        );
```
petorsong, line 43: - no real point in using `Promise.all` if you only have 1 promise in it
- the 3 updates can be done asynchronously since they should all be mutually exclusive

```suggestion
        queryInterface.bulkUpdate(
          'patients',
          { gender: 'F' },
          { gender: 'Female' },
          { transaction: t }
        ),
        queryInterface.bulkUpdate(
          'patients',
          { gender: 'M' },
          { gender: 'Male' },
          { transacton: t }
        ),
      ]).then(() =>
        queryInterface.changeColumn(
          'patients',
          'gender',
          {
            type: Sequelize.DataTypes.ENUM('M', 'F'),
          },
          {
            transaction: t,
          }
        )
      );
```

migrations/20200824194331-patient-gender.js
```diff
@@ -0,0 +1,82 @@
...
+            { gender: 'Male' },
+            { gender: 'M' },
+            { transaction: t }
+          ),
+          queryInterface.bulkUpdate(
+            'patients',
+            { gender: 'Female' },
+            { gender: 'F' },
+            { transaction: t }
+          )
```
petorsong, line 78: - only use `Promise.all` when relevant
- `.then()` takes a callback as an argument, so it's necessary to wrap our db call in a callback; i found [this was a great resource](https://javascript.info/promise-chaining) if you want to learn more about promise chaining

```suggestion
      return queryInterface
        .changeColumn(
          'patients',
          'gender',
          {
            type: Sequelize.DataTypes.STRING,
          },
          {
            transaction: t,
          }
        )
        .then(() =>
          queryInterface.sequelize.query('DROP TYPE enum_patients_gender;', {
            transaction: t,
          })
        )
        .then(() =>
          Promise.all([
            queryInterface.bulkUpdate(
              'patients',
              { gender: 'Male' },
              { gender: 'M' },
              { transaction: t }
            ),
            queryInterface.bulkUpdate(
              'patients',
              { gender: 'Female' },
              { gender: 'F' },
              { transaction: t }
            ),
          ])
```

PR #63, opened 2020-09-08
Fixing restoration resolvers

aaronabraham311:
This PR addresses two tickets:
[Ticket 1](https://www.notion.so/uwblueprintexecs/396370c5e09f4c28bc7f455a7e84c908?v=3751aa62a2594c74b7e0ecb363304c24&p=a2659349008a4914af16de78e596ba6b)
[Ticket 2](https://www.notion.so/uwblueprintexecs/396370c5e09f4c28bc7f455a7e84c908?v=3751aa62a2594c74b7e0ecb363304c24&p=54566102e29f49b3aeaca7a4d50055d5)

<b>Before reviewing this PR, please review the following documentation:</b>
1. [Paranoid tables](https://www.notion.so/uwblueprintexecs/6a0ad062ce504acc90fa25cb4d2c731f?v=4844a823da724b7ebf2c442a1ee1279f&p=4fabc50d4a9b4c969c96d7ae5bc19cfc)
2. [Deletion & restoration decisions](https://www.notion.so/uwblueprintexecs/6a0ad062ce504acc90fa25cb4d2c731f?v=4844a823da724b7ebf2c442a1ee1279f&p=a0701ccd6ca64c01898f09cd61ccfb8b)

## Changes
- Re-added `restoreEvent` resolver which exclusively restores the event and cascade restore any associated entities (not including resources)
- Changed `restoreAmbulance` and `restoreHospital` mutations so that it exclusively restores the entity and not the event-resource association
- Added error checking for all restoration resolvers

## Testing
Please perform the following tests:
1. Delete an event, confirm that the event is not returned in a query, restore an event and confirm that the event is now present in the query
2. Delete an event and confirm the following through queries:
    1. All associated collection points are deleted
    2. All associated patients are deleted
    Once confirmed, restore the event and confirm that associated collection points and patients are also restored
3. Navigate to the database and query `eventAmbulance` and `eventHospital` to determine the associations between events and ambulances. Then, delete an event, check that the associations are deleted in the `eventAmbulance` and `eventHospital` tables, restore the event, and check that the associations were NOT restored
4. Delete an ambulance, check the `eventAmbulance` table to confirm associations are deleted, restore the ambulance and check that the associations in the `eventAmbulance` table are NOT restored
5. Perform test case 5 using hospitals and the `eventHospital` junction table
6. Make sure that the association between events and hospitals/ambulance is NOT deleted when hospitals/ambulances already have patients. Perform the following steps:
    1. Perform `updatePatient` and add some patient to an ambulance/hospital
    2. Add the ambulance/hospital to an event by performing `addAmbulancesToEvent` or `addHospitalsToEvent`
    3. Try deleting the ambulance/hospital. An error should pop up
    4. Ensure that that the association is not deleted by querying `events` again and ensure that the results have not changed

Here is a GIF of me performing test 3 on for `eventAmbulance`:
Kapture 2020-09-11 at 12 01 37

### DIFF
--- a/resolvers/ambulance.js
+++ b/resolvers/ambulance.js
@@ -55,27 +55,29 @@ const ambulanceResolvers = {
       return db.ambulance.findByPk(args.id);
     },
     deleteAmbulance: async (parent, args) => {
-      await Promise.all([
-        db.patient
-          .count({
-            where: {
-              ambulanceId: args.id,
-            },
-          })
-          .then((count) => {
-            if (count > 0) {
-              throw new Error(
-                'Deletion failed; there are associated patients for ambulance ID: ' +
-                  args.id
-              );
-            }
-          }),
-        db.eventAmbulances.destroy({
+      await db.patient
+        .count({
           where: {
             ambulanceId: args.id,
           },
-        }),
-      ]);
+        })
+        .then((count) => {
+          if (count > 0) {
+            throw new Error(
+              'Deletion failed; there are associated patients for ambulance ID: ' +
+                args.id
+            );
+          }
+        })
+        .catch((error) => {
+          throw error;
+        });
+
+      await db.eventAmbulances.destroy({
+        where: {
+          ambulanceId: args.id,
+        },
+      });
 
       return db.ambulance.destroy({
         where: {
--- a/resolvers/ambulance.js
+++ b/resolvers/ambulance.js
@@ -45,13 +45,13 @@ const ambulanceResolvers = {
           return db.ambulance.findByPk(args.id);
         }),
     restoreAmbulance: async (parent, args) => {
-      await db.ambulance.restore({
-        where: {
-          id: args.id,
-        },
-      });
-
-      return db.ambulance.findByPk(args.id);
+      return db.ambulance
+        .restore({
+          where: {
+            id: args.id,
+          },
+        })
+        .then(() => db.ambulance.findByPk(args.id));
     },
     deleteAmbulance: async (parent, args) => {
       await Promise.all([
--- a/resolvers/ambulance.js
+++ b/resolvers/ambulance.js
@@ -45,39 +45,11 @@ const ambulanceResolvers = {
           return db.ambulance.findByPk(args.id);
         }),
     restoreAmbulance: async (parent, args) => {
-      await Promise.all([
-        db.ambulance.restore({
-          where: {
-            id: args.id,
-          },
-        }),
-        // Restoring event association if event also availiable
-        db.eventAmbulances
-          .findAll({
-            where: {
-              ambulanceId: args.id,
-            },
-            include: [
-              {
-                model: db.event,
-                required: true,
-              },
-            ],
-            paranoid: false,
-          })
-          .then((associatedEvents) =>
-            Promise.all(
-              associatedEvents.map((associatedEvent) =>
-                db.eventAmbulances.restore({
-                  where: {
-                    eventId: associatedEvent.eventId,
-                    ambulanceId: args.id,
-                  },
-                })
-              )
-            )
-          ),
-      ]);
+      await db.ambulance.restore({
+        where: {
+          id: args.id,
+        },
+      });
 
       return db.ambulance.findByPk(args.id);
     },
--- a/resolvers/ambulance.js
+++ b/resolvers/ambulance.js
@@ -27,7 +27,7 @@ const ambulanceResolvers = {
       db.ambulance.create({
         vehicleNumber: args.vehicleNumber,
       }),
-    updateAmbulance: (parent, args) => {
+    updateAmbulance: (parent, args) =>
       db.ambulance
         .update(
           {
@@ -44,12 +44,9 @@ const ambulanceResolvers = {
             throw new Error('Failed update for ambulance ID: ' + args.id);
           }
           return db.ambulance.findByPk(args.id);
-        });
-    },
+        }),
     restoreAmbulance: async (parent, args) => {
-      await validators.validateAmbulance(args.id, true).catch((error) => {
-        throw error;
-      });
+      await validators.validateAmbulance(args.id, true);
       await db.ambulance.restore({
         where: {
           id: args.id,
--- a/resolvers/ambulance.js
+++ b/resolvers/ambulance.js
@@ -1,6 +1,7 @@
 'use strict';
 
 const db = require('../models');
+const validators = require('../utils/validators');
 
 const ambulanceResolvers = {
   Query: {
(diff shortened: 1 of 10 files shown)


resolvers/ambulance.js
```diff
@@ -45,39 +45,11 @@ const ambulanceResolvers = {
           return db.ambulance.findByPk(args.id);
         }),
     restoreAmbulance: async (parent, args) => {
```
petorsong, line 47:
```restoreAmbulance: (parent, args) => {
  return db.ambulance.restore({
    where: {
      id: args.id,
    },
  }).then(() => db.ambulance.findByPk(args.id));
```

petorsong (commented):
test cases were solid - note that for cases 4 & 5, you need to use a hospital/ambulance that's not already associated with a patient as we don't allow the deletion to go through in that scenario

- if these mutation resolvers are becoming simplers, i prefer promise syntax over async/await
  - double check that errors thrown in the first promise get thrown and surfaced properly

resolvers/event.js
```diff
@@ -294,7 +294,16 @@ const eventResolvers = {
...
+    restoreEvent: async (parent, args) => {
+      await db.event.restore({
+        where: {
+          id: args.id,
+        },
+        individualHooks: true,
+      });
 
+      return db.event.findByPk(args.id);
+    },
```
petorsong, line 306:
```suggestion
    restoreEvent: (parent, args) =>
      db.event.restore({
        where: {
          id: args.id,
        },
        individualHooks: true,
      }).then(() => db.event.findByPk(args.id)),
```

resolvers/hospital.js
```diff
@@ -45,39 +45,11 @@ const hospitalResolvers = {
           return db.hospital.findByPk(args.id);
         }),
     restoreHospital: async (parent, args) => {
```
petorsong, line 47:
```
restoreHospital: (parent, args) => {
  return db.hospital.restore({
    where: {
      id: args.id,
    },
  }).then(() => db.hospital.findByPk(args.id));
```

petorsong (changes requested):

patrick-du (dismissed):
LGTM 👍

aaronabraham311:
I have added error checking for all restoration resolvers. Please check this as well!

resolvers/ambulance.js
```diff
@@ -45,41 +45,21 @@ const ambulanceResolvers = {
...
-                })
-              )
-            )
-          ),
-      ]);
-
-      return db.ambulance.findByPk(args.id);
+        })
+        .then(() => db.ambulance.findByPk(args.id));
     },
```
petorsong, line 63: since we're introducing a validation step, i figured it's a lot more readable if we just use `await` instead of promises
```
    restoreAmbulance: async (parent, args) => {
      await validators.validateAmbulance(args.id, true);
      await db.ambulance.restore({
        where: {
          id: args.id,
        },
      });

      return db.ambulance.findByPk(args.id);
    },
```

petorsong (changes requested):
i like the idea of adding this validation and it can be made a lot cleaner by modifying `validators.js` - i'll illustrate an example with ambulance but this change can be made for all of the restore resolvers you're touching in this PR

`validators.js`
```
...
  validateAmbulance: (
    ambulanceId,
    checkParanoid = false,
    errorMessage = 'Invalid ambulance ID: ' + ambulanceId
  ) => {
    const options = {};
    if (checkParanoid) {
      options.paranoid = false;
    }

    return db.ambulance.findByPk(ambulanceId, options).then((ambulance) => {
      if (!ambulance) {
        throw new Error(errorMessage);
      }
    });
  },
...
```

sarafong (dismissed):
lgtm!

jesskeepswimming (approved):
everything looks good to me!

however, I noticed that while deleting a hospital/ambulance with patients associated with it gives an error, the hospitals/ambulances still get deleted from the associated events. is this supposed to happen?

aaronabraham311:
Updated with a new test case thanks to @jesskeepswimming. Please check that before approving!

jesskeepswimming:
changes look good!

resolvers/ambulance.js
```diff
@@ -45,64 +46,38 @@ const ambulanceResolvers = {
...
+          if (count > 0) {
+            throw new Error(
+              'Deletion failed; there are associated patients for ambulance ID: ' +
+                args.id
+            );
+          }
+        })
+        .catch((error) => {
+          throw error;
+        });
```
petorsong, line 74: this one can go too
```suggestion
        });
```

resolvers/collectionPoint.js
```diff
@@ -61,15 +61,17 @@ const collectionPointResolvers = {
...
-        .restore({
-          where: {
-            id: args.id,
-          },
-          individualHooks: true,
-        })
-        .then(() => db.collectionPoint.findByPk(args.id)),
+    restoreCollectionPoint: async (parent, args) => {
+      await validators.validateCollectionPoint(args.id, true);
+
```
petorsong, line 66:
```suggestion
```

resolvers/hospital.js
```diff
@@ -45,64 +46,37 @@ const hospitalResolvers = {
...
+          if (count > 0) {
+            throw new Error(
+              'Deletion failed; there are associated patients for hospital ID: ' +
+                args.id
+            );
+          }
+        })
+        .catch((error) => {
+          throw error;
+        });
```
petorsong, line 74:
```suggestion
        });
```

petorsong (changes requested):
i was testing this out and noticed that the `users` table is not paranoid but has a `deletedAt` column - was this an oversight or intentional

aaronabraham311:
The `deletedAt` column is most likely an oversight. Should I create a new migration in this PR?

petorsong:
> 
> 
> The `deletedAt` column is most likely an oversight. Should I create a new migration in this PR?

as we have a `restoreUser` mutation an it makes sense for `users` to be paranoid, i just pushed a commit to make that change

petorsong (approved):

petorsong:
@jesskeepswimming @sarafong @patrick-du code looks good now so try the tests out again - we'll merge this PR in once we have 3 checkmarks

patrick-du (changes requested):
Test case 6, point 3 says "Try deleting the ambulance and patient. An error should pop up" - should it be ambulance and hospital? 

If so, all 6 test cases work as expected for me.

aaronabraham311:
Good catch. Updated the test case to ambulance/hospital.

patrick-du (approved):

PR #65, opened 2020-09-23
Added new roles for authorization

rickrm:
[Notion ticket](https://www.notion.so/uwblueprintexecs/Backend-Authorization-0f29dae7aad449e98ab43ee1e7f28316)

## Changes
- Added roles:
Commander
Supervisor
Dispatch

- Added role to model and schema
- Added name and displayName fields to role
- Added role migration with the populated data

## Testing
- Try migrating the new migrations and inspecting the database to see if the table has all the roles and their corresponding fields
- Also check Graph QL playground's schema to see if role's implementation is there

before opening PR
- [x] check notion ticket
- [x] run linter
- [x] go through file diff

filling out PR
- [x] descriptive title
- [x] fill out `Changes`
- [x] fill out `Testing`

after opening PR
- [x] move ticket to `In PR`
- [x] make sure build passes


models/role.js
```diff
@@ -0,0 +1,16 @@
+'use strict';
+
+module.exports = (sequelize, DataTypes) => {
+  const Role = sequelize.define(
```
petorsong, line 4: realized that this was inconsistent too - we should stick with lowercase
```suggestion
  const role = sequelize.define(
```

models/role.js
```diff
@@ -0,0 +1,16 @@
...
+    'role',
```
petorsong, line 5: should be `roles`
- we defined the table name to be `roles` in the migration
- table names should be plural as they store multiple entries of a single type of entity
- the model/export name (i.e. `const role = ...; ... export role;`) should be singular as we generally work with a single instance of an entity/model in code

hopefully that helps clear it up some
```suggestion
    'roles',
```

models/role.js
```diff
@@ -0,0 +1,16 @@
...
+    {
+      name: DataTypes.STRING,
+      displayName: DataTypes.STRING,
+    },
+    {
+      timestamps: false,
```
petorsong, line 11: good catch - normally we would want this, but for this `roles` table this is not necessary and also we don't actually want users to be able to edit entries in it, so i'm basically saying this is good lol

migrations/20200923012433-setup-roles.js
```diff
@@ -0,0 +1,56 @@
...
+            { transaction: t }
+          )
+        )
+    );
+  },
+
+  down: (queryInterface) => {
+    return queryInterface.sequelize.transaction((t) =>
+      queryInterface.dropTable('roles', { transaction: t })
+    );
```
petorsong, line 54: since there's only one (1) sequelize function being called here, there's no need for transactions
```suggestion
    return queryInterface.dropTable('roles');
```

petorsong (changes requested):
nice job! pointed out some minor refactoring but otherwise looking good

our sequelize models have been mad inconsistent - i have a ticket here to address that, feel free to add to it if you find more stuff you want dealt with
https://www.notion.so/uwblueprintexecs/model-inconsistencies-da7572ea79964c05a9313c19dc82eb80

petorsong (changes requested):
friendly reminder that my PR suggestions aren't meant to be comprehensive - you still need to go through the motions to make the changes locally and test that the changes are working correctly

see linter for some minor issues that need fixing

rickrm:
> friendly reminder that my PR suggestions aren't meant to be comprehensive - you still need to go through the motions to make the changes locally and test that the changes are working correctly
> 
> see linter for some minor issues that need fixing

Sorry, will remember to do it locally next time.

rickrm:
Checked the changes locally and linted.

petorsong (dismissed):

petorsong (approved):
reverted changes will be on a different PR

sarafong (approved):

PR #66, opened 2020-09-27
No duplicate patient barcodes

patrick-du:
[Notion ticket](https://www.notion.so/uwblueprintexecs/no-duplicate-barcodes-d0cd44a9ce9e4812bb250011edad50f7)

## Changes
- Added migration to:
    - Update existing null barcode values to be the patient ID
    - Update `barcodeValue` column in patient table to have a `NOT NULL` constraint
- Note that on `down` method, removing the constraint must occur first in order to reverse to null barcode values
- Updated GraphQL schema for patient by adding `!` to make `barcodeValue` non-nullable

## Testing
In `~/paramedics-shared/paramedics-web`, open two terminals
- To execute a bash shell on the container, run `pmc-s`
- To access the db, run `pmc-db`
- Prior to running migration, create a patient with no barcode value (null) and check DB to ensure patient exists
- Check the migration status with `sequelize db:migrate:status` in the bash shell
- Run the migration with `sequelize db:migrate`

- [ ] Test 1: Check DB to ensure patient now has a barcodeValue
- [ ] Test 2: Create a patient without a barcode value via the DB - this should produce an error: `ERROR:  null value in column "barcodeValue" violates not-null constraint`
- [ ] Test 3: Create a patient with a barcode value via DB - this should work
- [ ] Test 4: Create a patient without a barcode value via GraphQL Playground - this should produce an error with the message `"Field \"addPatient\" argument \"barcodeValue\" of type \"String!\" is required, but it was not provided."`
- [ ] Test 5: Create a patient with a barcode value via GraphQL Playground - this should work
- [ ] Test 6: Create a patient through frontend by scanning or manually entering the barcode (use the same barcode as an existing patient) - this should route you to edit the existing patient
- [ ] Test 7: Create a patient with a new (not existing) barcode value through the frontend - this should work

before opening PR
- [x] check notion ticket
- [x] run linter
- [x] go through file diff

filling out PR
- [x] descriptive title
- [x] fill out `Changes`
- [x] fill out `Testing`

after opening PR
- [x] move ticket to `In PR`
- [x] make sure build passes

### DIFF
--- a/migrations/20200927132044-update-patient-barcode-non-nullable.js
+++ b/migrations/20200927132044-update-patient-barcode-non-nullable.js
@@ -1,0 +1,33 @@
+'use strict';
+
+module.exports = {
+  up: (queryInterface) => {
+    return queryInterface.sequelize.transaction((t) => {
+      return Promise.all([
+        queryInterface.sequelize.query(
+          `UPDATE patients SET "barcodeValue" = "id" WHERE "barcodeValue" IS NULL`,
+          { transaction: t }
+        ),
+        queryInterface.sequelize.query(
+          'ALTER TABLE patients ALTER COLUMN "barcodeValue" TYPE "varchar", ALTER COLUMN "barcodeValue" SET NOT NULL',
+          { transaction: t }
+        ),
+      ]);
+    });
+  },
+
+  down: (queryInterface) => {
+    return queryInterface.sequelize.transaction((t) => {
+      return Promise.all([
+        queryInterface.sequelize.query(
+          'ALTER TABLE patients ALTER COLUMN "barcodeValue" TYPE "varchar", ALTER COLUMN "barcodeValue" DROP NOT NULL',
+          { transaction: t }
+        ),
+        queryInterface.sequelize.query(
+          `UPDATE patients SET "barcodeValue" = NULL WHERE "barcodeValue" = "id"::varchar`,
+          { transaction: t }
+        ),
+      ]);
+    });
+  },
+};
--- a/migrations/20200927132044-update-patient-barcode-non-nullable.js
+++ b/migrations/20200927132044-update-patient-barcode-non-nullable.js
@@ -1,41 +1,24 @@
 'use strict';
 
 module.exports = {
-  up: (queryInterface) => {
-    return queryInterface.sequelize.transaction((t) => {
-      return Promise.all([
-        queryInterface.sequelize.query(
-          `UPDATE patients SET "barcodeValue" = "id" WHERE "barcodeValue" IS NULL`,
-          { transaction: t }
-        ),
-        queryInterface.sequelize.query(
-          `ALTER TABLE patients ALTER COLUMN "barcodeValue" TYPE "varchar", ALTER COLUMN "barcodeValue" SET NOT NULL`,
-          { transaction: t }
-        ),
-        queryInterface.sequelize.query(
-          `ALTER TABLE patients ADD CONSTRAINT unique_barcode UNIQUE ("barcodeValue")`,
-          { transaction: t }
-        ),
-      ]);
+  up: async (queryInterface, Sequelize) => {
+    await queryInterface.sequelize.query(
+      `UPDATE patients SET "barcodeValue" = "id" WHERE "barcodeValue" IS NULL`
+    );
+    return queryInterface.changeColumn('patients', 'barcodeValue', {
+      type: Sequelize.STRING,
+      allowNull: false,
+      unique: true,
     });
   },
-
-  down: (queryInterface) => {
-    return queryInterface.sequelize.transaction((t) => {
-      return Promise.all([
-        queryInterface.sequelize.query(
-          `ALTER TABLE patients DROP CONSTRAINT "unique_barcode"`,
-          { transaction: t }
-        ),
-        queryInterface.sequelize.query(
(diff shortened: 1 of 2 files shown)


schema/patient.js
```diff
@@ -67,7 +67,7 @@ const patientSchema = `
     gender: Gender,
     age: Int,
     runNumber: Int,
-    barcodeValue: String,
+    barcodeValue: String!,
```
petorsong, line 70: i noticed that the `addPatient` mutation has `triageLevel` as nullable (line 37) - we should have the required fields for that and this `type Patient` to match (i.e. `barcodeValue`, `collectionPoint`, `triageLevel` should be required/non-nullable for both the type and the `addPatient` mutation

(not exactly in the scope of your ticket but i figure it's a quick minor change that's easy to make here)

migrations/20200927132044-update-patient-barcode-non-nullable.js
```diff
@@ -0,0 +1,33 @@
...
+    return queryInterface.sequelize.transaction((t) => {
+      return Promise.all([
+        queryInterface.sequelize.query(
+          `UPDATE patients SET "barcodeValue" = "id" WHERE "barcodeValue" IS NULL`,
+          { transaction: t }
+        ),
+        queryInterface.sequelize.query(
+          'ALTER TABLE patients ALTER COLUMN "barcodeValue" TYPE "varchar", ALTER COLUMN "barcodeValue" SET NOT NULL',
+          { transaction: t }
+        ),
```
petorsong, line 14: [there is an existing sequelize method that can alter a column without a raw query](https://sequelize.org/master/class/lib/dialects/abstract/query-interface.js~QueryInterface.html#instance-method-changeColumn) - is it possible to use this instead? this seems straightforward enough that we can use a sequelize method, which is preferable as they _should_ be more consistent than raw queries (same deal for the equivalent `down` query)

migrations/20200927132044-update-patient-barcode-non-nullable.js
```diff
@@ -0,0 +1,33 @@
...
+        queryInterface.sequelize.query(
+          'ALTER TABLE patients ALTER COLUMN "barcodeValue" TYPE "varchar", ALTER COLUMN "barcodeValue" DROP NOT NULL',
+          { transaction: t }
+        ),
+        queryInterface.sequelize.query(
+          `UPDATE patients SET "barcodeValue" = NULL WHERE "barcodeValue" = "id"::varchar`,
+          { transaction: t }
+        ),
+      ]);
+    });
```
petorsong, line 31: `Promise.all` runs everything in its array concurrently - this means that order is not guaranteed

to ensure that the order is followed, we can use `await`/`async` syntax and do something like this (please test as i just threw this together quickly) - this can also be done in the `up` mutation for the same purpose
```suggestion
    return queryInterface.sequelize.transaction(async (t) => {
      await queryInterface.sequelize.query(
          'ALTER TABLE patients ALTER COLUMN "barcodeValue" TYPE "varchar", ALTER COLUMN "barcodeValue" DROP NOT NULL',
          { transaction: t }
        );
      return queryInterface.sequelize.query(
          `UPDATE patients SET "barcodeValue" = NULL WHERE "barcodeValue" = "id"::varchar`,
          { transaction: t }
        );
    });
```

petorsong (changes requested):
thanks for being thorough with the testing steps! one way to skip a lot of the switching back and forth stuff at the beginning (specifically with new migrations) is to switch to the branch then start your container - migrations are automatically run when the container starts

**i also noticed that we actually aren't enforcing `barcodeValue` with a unique constraint - ideally that would also be included in this PR**

migrations/20200927132044-update-patient-barcode-non-nullable.js
```diff
@@ -0,0 +1,27 @@
...
+  },
+  down: async (queryInterface, Sequelize) => {
+    await Promise.all([
+      queryInterface.changeColumn('patients', 'barcodeValue', {
+        type: Sequelize.STRING,
+        allowNull: true,
+        unique: false,
+      }),
+      queryInterface.removeConstraint('patients', 'patients_barcodeValue_key'),
```
petorsong, line 21: i noticed that sequelize is kinda silly and didn't clean up the constraints when repeatedly running and undoing the migration, so i pushed this commit to fix that (so we hopefully make it faster for this PR to get merged)
image

migrations/20200927132044-update-patient-barcode-non-nullable.js
```diff
@@ -0,0 +1,27 @@
...
+      `UPDATE patients SET "barcodeValue" = "id" WHERE "barcodeValue" IS NULL`
+    );
+    return queryInterface.changeColumn('patients', 'barcodeValue', {
+      type: Sequelize.STRING,
+      allowNull: false,
+      unique: true,
+    });
+  },
+  down: async (queryInterface, Sequelize) => {
+    await Promise.all([
```
petorsong, line 15: since it shouldn't matter the order that the constraint is removed and the column changed, so we can send run them consecutively to gain a (tiny) performance boost

petorsong (approved):
pushed a change to fix an issue i saw, take a look at my comments for context but otherwise i'm good to merge 🥡

carelynntsai (approved):
great job!! 🍕

PR #67, opened 2020-09-27
Update all delete mutations to return entity id

wongcmichelle:
[Notion ticket](https://www.notion.so/uwblueprintexecs/delete-and-edit-mutation-cleanup-98ad81733016484b928909fb184e7360)

## Changes
- Update following delete mutations to return the deleted ID
    - ambulance
    - CCP (collectionPoint)
    - event
    - hospital
    - locationPin
    - patient
    - user

## Testing
**Ambulance**
Delete Successfully
image
image
Delete Error
- Error deleting ambulance with associated patients
image
- Error deleting ambulance with invalid ID
image

**CCP**
Delete Successfully
image
image
Delete Error
- Invalid ID
image

**Event**
Delete Successfully
image
image
Delete Error
- Invalid ID
image

**Hospital**
Delete Successfully
image
image

Delete Error
- Error deleting hospital with associated patients
image
- Error deleting hospital with invalid ID
image

**Location Pin**
Delete Successfully
image
image
Delete Error
- Invalid ID
image

**Patient**
Delete Successfully
image
image
Delete Error
- Invalid ID
image

**User**
Delete Successfully
image
image
Delete Error
- Invalid ID
image

before opening PR
- [x] check notion ticket
- [x] run linter
- [x] go through file diff

filling out PR
- [x] descriptive title
- [x] fill out `Changes`
- [x] fill out `Testing`

after opening PR
- [x] move ticket to `In PR`
- [x] make sure build passes

### DIFF
--- a/resolvers/ambulance.js
+++ b/resolvers/ambulance.js
@@ -81,35 +81,41 @@ const ambulanceResolvers = {
 
       return db.ambulance.findByPk(args.id);
     },
-    deleteAmbulance: async (parent, args) => {
-      await Promise.all([
-        db.patient
-          .count({
+    deleteAmbulance: async (parent, args) =>
+      db.patient
+        .count({
+          where: {
+            ambulanceId: args.id,
+          },
+        })
+        .then((count) => {
+          if (count > 0) {
+            throw new Error(
+              'Deletion failed; there are associated patients for ambulance ID: ' +
+                args.id
+            );
+          }
+        })
+        .then(() =>
+          db.eventAmbulances.destroy({
             where: {
               ambulanceId: args.id,
             },
           })
-          .then((count) => {
-            if (count > 0) {
-              throw new Error(
-                'Deletion failed; there are associated patients for ambulance ID: ' +
-                  args.id
-              );
-            }
-          }),
-        db.eventAmbulances.destroy({
-          where: {
-            ambulanceId: args.id,
-          },
+        )
+        .then(() =>
+          db.ambulance.destroy({
+            where: {
+              id: args.id,
+            },
+          })
+        )
+        .then((isDeleted) => {
+          if (isDeleted === 1) {
+            return args.id;
+          }
+          throw new Error('Deletion failed for ambulance ID: ' + args.id);
         }),
-      ]);
-
-      return db.ambulance.destroy({
-        where: {
-          id: args.id,
-        },
-      });
-    },
   },
 };
 
--- a/resolvers/event.js
+++ b/resolvers/event.js
@@ -295,10 +295,8 @@ const eventResolvers = {
       });
     },
 
-    deleteEvent: async (parent, args) => {
-      // Return status for destroy
-      // 1 for successful deletion, 0 otherwise
-      await Promise.all([
+    deleteEvent: async (parent, args) =>
+      Promise.all([
         db.eventAmbulances.destroy({
           where: {
             eventId: args.id,
@@ -309,15 +307,21 @@ const eventResolvers = {
             eventId: args.id,
           },
         }),
-      ]);
-
-      return db.event.destroy({
-        where: {
-          id: args.id,
-        },
-        individualHooks: true,
-      });
-    },
+      ])
+        .then(() =>
+          db.event.destroy({
+            where: {
+              id: args.id,
+            },
+            individualHooks: true,
+          })
+        )
+        .then((isDeleted) => {
+          if (isDeleted === 1) {
+            return args.id;
+          }
+          throw new Error('Deletion failed for event ID: ' + args.id);
+        }),
   },
 };
 
--- a/resolvers/hospital.js
+++ b/resolvers/hospital.js
@@ -81,8 +81,8 @@ const hospitalResolvers = {
 
       return db.hospital.findByPk(args.id);
     },
-    deleteHospital: async (parent, args) => {
-      await Promise.all([
+    deleteHospital: async (parent, args) =>
+      Promise.all([
         db.patient
           .count({
             where: {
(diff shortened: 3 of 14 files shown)


resolvers/ambulance.js
```diff
@@ -68,20 +68,27 @@ const ambulanceResolvers = {
...
+              id: args.id,
+            },
+          })
+        )
+        .then((isDeleted) => {
+          if (isDeleted === 1) {
+            return args.id;
+          }
+          throw new Error('Deletion failed for ambulance ID: ' + args.id);
+        }),
```
wongcmichelle, line 91: `destroy` returns a Promise that always gets resolved with `1` on success or `0` otherwise (i.e. the promise is always “fulfilled” and won’t trigger a `catch` statement event when the deleting failed)

resolvers/event.js
```diff
@@ -304,10 +304,8 @@ const eventResolvers = {
       });
       return db.event.findByPk(args.id);
     },
-    deleteEvent: async (parent, args) => {
-      // Return status for destroy
-      // 1 for successful deletion, 0 otherwise
-      await Promise.all([
+    deleteEvent: async (parent, args) =>
+      Promise.all([
         db.eventAmbulances.destroy({
```
wongcmichelle, line 309: destroy returns a Promise that always gets resolved with 1 on success or 0 otherwise (i.e. the promise is always “fulfilled” and won’t trigger a catch statement event when the deleting failed).

Note this means that there's no "error handling" for `db.eventAmbulances.destroy` and `db.eventHospitals.destroy`. We can implement it but I couldn't find a clean way to do it.

resolvers/ambulance.js
```diff
@@ -68,20 +68,27 @@ const ambulanceResolvers = {
...
+            where: {
+              id: args.id,
+            },
+          })
+        )
+        .then((isDeleted) => {
+          if (isDeleted === 1) {
+            return args.id;
+          }
+          throw new Error('Deletion failed for ambulance ID: ' + args.id);
```
wongcmichelle, line 90: We have no way of knowing why it failed at this point, so that's why the error message is pretty general.

resolvers/hospital.js
```diff
@@ -68,19 +68,27 @@ const hospitalResolvers = {
...
-      });
-    },
+        })
+        .then(() =>
+          db.eventHospitals.destroy({
+            where: {
+              hospitalId: args.id,
+            },
+          })
+        )
```
wongcmichelle, line 78: destroy returns a Promise that always gets resolved with 1 on success or 0 otherwise (i.e. the promise is always “fulfilled” and won’t trigger a catch statement event when the deleting failed).

Note this means that there's no "error handling" for `db.eventHospitals.destroy` here.

resolvers/locationPin.js
```diff
@@ -69,13 +69,18 @@ const locationPinResolvers = {
...
+          where: {
+            id: args.id,
+          },
+        })
+        .then((isDeleted) => {
+          if (isDeleted === 1) {
+            return args.id;
+          }
+          throw new Error('Deletion failed for location pin ID: ' + args.id);
+        }),
```
petorsong, line 83: is there a way to generalize this and pull it out like we did for `utils/validators`? (i know it might not work well with the resource associations but this is a little repetitive)

resolvers/ambulance.js
```diff
@@ -54,8 +54,8 @@ const ambulanceResolvers = {
       });
       return db.ambulance.findByPk(args.id);
     },
-    deleteAmbulance: async (parent, args) => {
-      await db.patient
+    deleteAmbulance: async (parent, args) =>
+      db.patient
```
petorsong, line 58: i'm not a fan of promise chaining especially this many times - i know i've leaned towards it before but i feel like there's been issues with readability and swallowing errors thrown inside promises; to be clear, i think one promise + one chain is reasonable, but beyond that it gets kinda messy, especially if there's not a whole lot happening inside them

i'm liking `async`/`await` more now, as it's more intuitive (maybe just for me) and easier to deal with errors - i've been  pushing for that with new devs (e.g. https://github.com/uwblueprint/paramedics-web/pull/66)

just wanted to thoughts from you and @aaronabraham311 on what pattern we should move forward with

petorsong (commented):
pretty straightforward, just 2 comments i'd like addressed!

resolvers/locationPin.js
wongcmichelle: could take this part out `(isDeleted) => {
          if (isDeleted === 1) {
            return args.id;
          }
          throw new Error('Deletion failed for location pin ID: ' + args.id);
        }`

and define something like `successCallback = (isDeleted, entityTypeLabel) => {
          if (isDeleted === 1) {
            return args.id;
          }
          throw new Error(`Deletion failed for ${entityTypeLabel} ID: ${args.id}`);
        }`

resolvers/ambulance.js
wongcmichelle: I personally think it's cleaner to use promise chaining here if we're using the result returned by these promises. Since like I mentioned, `try` and `catch` wouldn't do anything for us because `destroy` always results in a fulfilled promise, but it's just a matter of whether its return value is `0` or `1`.

wongcmichelle: I don't think it's about whether you consistently use async/await or promise chaining, since they each have their uses https://medium.com/better-programming/should-i-use-promises-or-async-await-126ab5c98789

resolvers/locationPin.js
petorsong: Your call if you want to do that refactoring - our usage of `destroy` is not super consistent/clean so I'm ok if we leave it this way

resolvers/ambulance.js
petorsong: Thanks for the link, that does help with clarifying between the two options - in that case I'm fine with keeping this the way it is

aaronabraham311 (approved):
Awesome work, lgtm!

petorsong (approved):

resolvers/locationPin.js
wongcmichelle: @petorsong do you want this refactoring before i merge the PR?

petorsong: Nah it's fine, I'm good with merging as is

PR #68, opened 2020-10-07
Returning patient object from deletion resolver

aaronabraham311:
[Notion ticket](https://www.notion.so/uwblueprintexecs/9224caeb21ce4350b8a1dc378d393323?v=43f44ac04d734394b7f7edc636c523b4)

## Changes
- Returning an entire patient object from the deletion resolver rather than only it's ID
    - Motivation: we need to update the cache on the frontend, which requires conforming to the 'edit cache' pattern where we need to return both the ID as well as the fields that we have changed. This allows us to do the latter
## Screenshots
Querying the status from a deleted patient: 
Screen Shot 2020-10-06 at 8 31 13 PM
## Testing
1. Make sure that you can query <b>all</b> fields from the returned patient entity
2. Create/update a patient and perform test 1 again

before opening PR
- [ ] check notion ticket
- [ ] run linter
- [ x ] go through file diff

filling out PR
- [x] descriptive title
- [ ] update `notion ticket` link
- [x] fill out `Changes`
- [x] fill out `Testing`

after opening PR
- [ ] link PR to notion ticket
- [ ] move ticket to `In PR`
- [x] make sure build passes

### DIFF
--- a/resolvers/patient.js
+++ b/resolvers/patient.js
@@ -90,8 +90,8 @@ const patientResolvers = {
       return db.patient.findByPk(args.id);
     },
     // This is a user delete of a patient, where the status is updated. A system delete happens if a CCP with associated patients is deleted
-    deletePatient: (parent, args) =>
-      db.patient
+    deletePatient: async (parent, args) => {
+      await db.patient
         .update(
           {
             status: 'DELETED',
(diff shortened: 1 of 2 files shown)


resolvers/patient.js
```diff
@@ -105,7 +105,9 @@ const patientResolvers = {
             return args.id;
```
petorsong, line 105: shouldn't this be the line that gets changed? technically we should never reach line 109 as we would always return or `throw` beforehand

petorsong (approved):
pretty straightforward 🤸

wongcmichelle (approved):
LGTM 🚢

PR #69, opened 2020-10-11
Added role validators to resolvers

rickrm:
[Notion ticket](https://www.notion.so/uwblueprintexecs/Backend-Authorization-0f29dae7aad449e98ab43ee1e7f28316)

## Changes

- Added role validators to:
Ambulances,
Collection Points,
Events,
Hospitals,
Location Pins,
Patients,
Users,

- Added the implementation of the role validator to `validator.js`
- Added demoRole for testing
-  Added enumerations for roles:
COMMANDER,
SUPERVISOR,
DISPATCH,

## Testing
- Test by changing the role in`demoRole` in `validator.js`
- Check each resolver through GraphQL to see if it catches roles that do not have sufficient permissions
- Check that each resolver worked if the demo role had sufficient permission 
- Made sure the role argument for each resolver matches the spreadsheet permissions

While testing, please consult [this](https://docs.google.com/spreadsheets/d/1JTQlflqYCW8p78ceiyYTEJxTe1UuTgLT7OhHtPeyyV8/edit#gid=1413489646) spreadsheet regarding the permissions associated with each role.

before opening PR
- [x] check notion ticket
- [x] run linter
- [x] go through file diff

filling out PR
- [x] descriptive title
- [x] fill out `Changes`
- [x] fill out `Testing`

after opening PR
- [x] move ticket to `In PR`
- [x] make sure build passes

### DIFF
--- a/resolvers/ambulance.js
+++ b/resolvers/ambulance.js
@@ -17,7 +17,7 @@ const ambulanceResolvers = {
     },
     ambulance: (parent, args) => {
       validators.validateRole(['COMMANDER']);
-      db.ambulance.findByPk(args.id, {
+      return db.ambulance.findByPk(args.id, {
         include: [
           {
             model: db.event,
@@ -29,7 +29,7 @@ const ambulanceResolvers = {
   Mutation: {
     addAmbulance: (parent, args) => {
       validators.validateRole(['COMMANDER']);
-      db.ambulance.create({
+      return db.ambulance.create({
         vehicleNumber: args.vehicleNumber,
       });
     },
--- a/resolvers/ambulance.js
+++ b/resolvers/ambulance.js
@@ -35,9 +35,13 @@ const ambulanceResolvers = {
     },
     updateAmbulance: (parent, args) => {
       validators.validateRole(['COMMANDER']);
-      db.ambulance
+      return db.ambulance
         .update({
           vehicleNumber: args.vehicleNumber,
+        }, {
+          where: {
+            id: args.id,
+          },
         })
         .then((rowsAffected) => {
           if (rowsAffected[0] === 0) {
@@ -58,7 +62,7 @@ const ambulanceResolvers = {
     },
     deleteAmbulance: async (parent, args) => {
       validators.validateRole(['COMMANDER']);
-      db.patient
+      return db.patient
         .count({
           where: {
             ambulanceId: args.id,
--- a/resolvers/ambulance.js
+++ b/resolvers/ambulance.js
@@ -1,12 +1,13 @@
 'use strict';
 
 const db = require('../models');
+const { Roles } = require('../utils/enum');
 const validators = require('../utils/validators');
 
 const ambulanceResolvers = {
   Query: {
     ambulances: () => {
-      validators.validateRole(['COMMANDER']);
+      validators.validateRole([Roles.COMMANDER], validators.demoRole);
       return db.ambulance.findAll({
         include: [
           {
@@ -16,7 +17,7 @@ const ambulanceResolvers = {
       });
     },
     ambulance: (parent, args) => {
-      validators.validateRole(['COMMANDER']);
+      validators.validateRole([Roles.COMMANDER], validators.demoRole);
       return db.ambulance.findByPk(args.id, {
         include: [
           {
@@ -28,13 +29,13 @@ const ambulanceResolvers = {
   },
   Mutation: {
     addAmbulance: (parent, args) => {
-      validators.validateRole(['COMMANDER']);
+      validators.validateRole([Roles.COMMANDER], validators.demoRole);
       return db.ambulance.create({
         vehicleNumber: args.vehicleNumber,
       });
     },
     updateAmbulance: (parent, args) => {
-      validators.validateRole(['COMMANDER']);
+      validators.validateRole([Roles.COMMANDER], validators.demoRole);
       return db.ambulance
         .update(
           {
@@ -54,7 +55,7 @@ const ambulanceResolvers = {
         });
     },
     restoreAmbulance: async (parent, args) => {
-      validators.validateRole(['COMMANDER']);
+      validators.validateRole([Roles.COMMANDER], validators.demoRole);
       await validators.validateAmbulance(args.id, true);
       await db.ambulance.restore({
         where: {
@@ -64,7 +65,7 @@ const ambulanceResolvers = {
       return db.ambulance.findByPk(args.id);
     },
     deleteAmbulance: async (parent, args) => {
-      validators.validateRole(['COMMANDER']);
+      validators.validateRole([Roles.COMMANDER], validators.demoRole);
       return db.patient
         .count({
           where: {
--- a/resolvers/ambulance.js
+++ b/resolvers/ambulance.js
@@ -7,10 +7,7 @@ const validators = require('../utils/validators');
 const ambulanceResolvers = {
   Query: {
     ambulances: () => {
-      validators.validateRole(
-        Object.keys(Roles),
-        validators.demoRole
-      );
+      validators.validateRole(Object.values(Roles), validators.demoRole);
       return db.ambulance.findAll({
         include: [
           {
@@ -20,10 +17,7 @@ const ambulanceResolvers = {
       });
     },
     ambulance: (parent, args) => {
-      validators.validateRole(
-        Object.keys(Roles),
-        validators.demoRole
-      );
+      validators.validateRole(Object.values(Roles), validators.demoRole);
       return db.ambulance.findByPk(args.id, {
         include: [
           {
@@ -82,7 +76,7 @@ const ambulanceResolvers = {
           if (count > 0) {
             throw new Error(
               'Deletion failed; there are associated patients for ambulance ID: ' +
-              args.id
+                args.id
             );
           }
         })
--- a/resolvers/ambulance.js
+++ b/resolvers/ambulance.js
@@ -26,28 +26,26 @@ const ambulanceResolvers = {
       });
     },
   },
-    Mutation: {
-      addAmbulance: (parent, args) => {
-        validators.validateRole(['COMMANDER']);
-        db.ambulance.create({
+  Mutation: {
+    addAmbulance: (parent, args) => {
+      validators.validateRole(['COMMANDER']);
+      db.ambulance.create({
+        vehicleNumber: args.vehicleNumber,
+      });
+    },
+    updateAmbulance: (parent, args) => {
+      validators.validateRole(['COMMANDER']);
+      db.ambulance
+        .update({
           vehicleNumber: args.vehicleNumber,
-        });
-      },
-      updateAmbulance: (parent, args) => {
-        validators.validateRole(['COMMANDER']);
-        db.ambulance
-          .update(
-            {
-              vehicleNumber: args.vehicleNumber,
-            },
-        )
+        })
         .then((rowsAffected) => {
           if (rowsAffected[0] === 0) {
             throw new Error('Failed update for ambulance ID: ' + args.id);
           }
           return db.ambulance.findByPk(args.id);
-        })
-      },
+        });
+    },
     restoreAmbulance: async (parent, args) => {
       validators.validateRole(['COMMANDER']);
       await validators.validateAmbulance(args.id, true);
@@ -93,8 +91,8 @@ const ambulanceResolvers = {
             return args.id;
           }
           throw new Error('Deletion failed for ambulance ID: ' + args.id);
-        })
-    }
+        });
+    },
   },
 };
 exports.ambulanceResolvers = ambulanceResolvers;
--- a/resolvers/ambulance.js
+++ b/resolvers/ambulance.js
@@ -1,116 +1,126 @@
 'use strict';
 
 const db = require('../models');
+const { validateRole } = require('../utils/validators');
 
 const ambulanceResolvers = {
   Query: {
-    ambulances: () =>
-      db.ambulance.findAll({
+    ambulances: () => {
+      validateRole(['ADMIN', 'COMMANDER']);
+      return db.ambulance.findAll({
         include: [
           {
             model: db.event,
           },
         ],
-      }),
-    ambulance: (parent, args) =>
+      });
+    },
+    ambulance: (parent, args) => {
+      validateRole(['ADMIN', 'COMMANDER']);
       db.ambulance.findByPk(args.id, {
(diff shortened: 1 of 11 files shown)


utils/validators.js
```diff
@@ -111,4 +112,16 @@ module.exports = {
         }
       });
   },
+  validateRole: (role, errorMessage = 'Insufficient permission') => {
```
petorsong, line 115: this copy is a little too vague, see suggestion or anything along those lines
```suggestion
  validateRole: (role, errorMessage = 'You do not have sufficient permission to perform this action') => {
```

resolvers/event.js
```diff
@@ -5,8 +5,9 @@ const validators = require('../utils/validators');
 
 const eventResolvers = {
   Query: {
-    events: () =>
-      db.event.findAll({
+    events: () => {
+      validators.validateRole(['COMMANDER', 'SUPERVISOR', 'DISPATCH']);
```
petorsong, line 9: `role` is currently an array of strings, ideally we would want to work with enums here instead

if it doesn't make sense to make the change now, would need to be tackled with authentication by @JasonYG

resolvers/ambulance.js
```diff
@@ -5,30 +5,37 @@ const validators = require('../utils/validators');
 
 const ambulanceResolvers = {
   Query: {
-    ambulances: () =>
-      db.ambulance.findAll({
+    ambulances: () => {
+      validators.validateRole(['COMMANDER']);
```
petorsong, line 9: we want `validateRole` to take the user's current role, as it should not be accessible from within `validators` - this will require you to hard code a user role in every one of these resolvers, example below
```
...
// TODO: remove this after authentication is implemented
const userRole = 'COMMANDER';
...
validators.validateRole(userRole, ['COMMANDER']);
```

petorsong (changes requested):
also, please explain (in the PR description) how to test that these role permissions are correct for each query/mutation

utils/validators.js
rickrm: Ok will do

resolvers/event.js
rickrm: I'll discuss with Jason regarding this.

resolvers/ambulance.js
```diff
@@ -1,34 +1,42 @@
...
 const db = require('../models');
+const { Roles } = require('../utils/enum');
 const validators = require('../utils/validators');
 
 const ambulanceResolvers = {
   Query: {
-    ambulances: () =>
-      db.ambulance.findAll({
+    ambulances: () => {
+      validators.validateRole([Roles.COMMANDER], validators.demoRole);
```
petorsong, line 10:
```suggestion
      validators.validateRole([Roles.COMMANDER, Roles.SUPERVISOR, Roles.DISPATCH], validators.demoRole);
```

resolvers/ambulance.js
```diff
@@ -1,34 +1,42 @@
...
             model: db.event,
           },
         ],
-      }),
-    ambulance: (parent, args) =>
-      db.ambulance.findByPk(args.id, {
+      });
+    },
+    ambulance: (parent, args) => {
+      validators.validateRole([Roles.COMMANDER], validators.demoRole);
```
petorsong, line 20:
```suggestion
      validators.validateRole([Roles.COMMANDER, Roles.SUPERVISOR, Roles.DISPATCH], validators.demoRole);
```

resolvers/hospital.js
```diff
@@ -1,34 +1,42 @@
...
 const db = require('../models');
+const { Roles } = require('../utils/enum');
 const validators = require('../utils/validators');
 
 const hospitalResolvers = {
   Query: {
-    hospitals: () =>
-      db.hospital.findAll({
+    hospitals: () => {
+      validators.validateRole([Roles.COMMANDER], validators.demoRole);
```
petorsong, line 10:
```suggestion
      validators.validateRole([Roles.COMMANDER, Roles.SUPERVISOR, Roles.DISPATCH], validators.demoRole);
```

resolvers/hospital.js
```diff
@@ -1,34 +1,42 @@
...
             model: db.event,
           },
         ],
-      }),
-    hospital: (parent, args) =>
-      db.hospital.findByPk(args.id, {
+      });
+    },
+    hospital: (parent, args) => {
+      validators.validateRole([Roles.COMMANDER], validators.demoRole);
```
petorsong, line 20:
```suggestion
      validators.validateRole([Roles.COMMANDER, Roles.SUPERVISOR, Roles.DISPATCH], validators.demoRole);
```

resolvers/user.js
```diff
@@ -1,24 +1,34 @@
...
 const db = require('../models');
+const { Roles } = require('../utils/enum');
 const validators = require('../utils/validators');
 
 const userResolvers = {
   Query: {
-    users: () => db.user.findAll(),
-    user: (parent, args) => db.user.findByPk(args.id),
+    users: () => {
+      validators.validateRole([Roles.COMMANDER], validators.demoRole);
```
petorsong, line 10:
```suggestion
      validators.validateRole([Roles.COMMANDER, Roles.SUPERVISOR, Roles.DISPATCH], validators.demoRole);
```

resolvers/user.js
```diff
@@ -1,24 +1,34 @@
...
+      return db.user.findAll();
+    },
+    user: (parent, args) => {
+      validators.validateRole([Roles.COMMANDER], validators.demoRole);
```
petorsong, line 14:
```suggestion
      validators.validateRole([Roles.COMMANDER, Roles.SUPERVISOR, Roles.DISPATCH], validators.demoRole);
```

resolvers/locationPin.js
```diff
@@ -1,18 +1,36 @@
...
 
 const locationPinResolvers = {
   Query: {
-    pins: () => db.locationPins.findAll(),
-    pin: (parent, args) => db.locationPins.findByPk(args.id),
-    pinsForEvent: (parent, args) =>
-      db.locationPins.findAll({
+    pins: () => {
+      validators.validateRole(
+        [Roles.COMMANDER, Roles.SUPERVISOR],
```
petorsong, line 11:
```suggestion
        [Roles.COMMANDER, Roles.SUPERVISOR, Roles.DISPATCH],
```

resolvers/locationPin.js
```diff
@@ -1,18 +1,36 @@
...
+        validators.demoRole
+      );
+      return db.locationPins.findAll();
+    },
+    pin: (parent, args) => {
+      validators.validateRole(
+        [Roles.COMMANDER, Roles.SUPERVISOR],
```
petorsong, line 18:
```suggestion
        [Roles.COMMANDER, Roles.SUPERVISOR, Roles.DISPATCH],
```

resolvers/locationPin.js
```diff
@@ -1,18 +1,36 @@
...
+        validators.demoRole
+      );
+      return db.locationPins.findByPk(args.id);
+    },
+    pinsForEvent: (parent, args) => {
+      validators.validateRole(
+        [Roles.COMMANDER, Roles.SUPERVISOR],
```
petorsong, line 25:
```suggestion
        [Roles.COMMANDER, Roles.SUPERVISOR, Roles.DISPATCH],
```

resolvers/event.js
```diff
@@ -26,19 +36,23 @@ const eventResolvers = {
...
+      return db.event.findAll({
         where: {
           isActive: false,
         },
-      }),
+      });
+    },
   },
   Event: {
     createdBy: (parent) => db.user.findByPk(parent.createdBy),
```
petorsong, line 51: let's just put a generic validator for all three roles on this query too

resolvers/collectionPoint.js
```diff
@@ -1,26 +1,46 @@
...
+        [Roles.COMMANDER, Roles.SUPERVISOR, Roles.DISPATCH],
+        validators.demoRole
+      );
+      return db.collectionPoint.findAll({ where: { eventId: args.eventId } });
+    },
   },
-
   collectionPoint: {
     eventId: (parent) => db.event.findByPk(parent.eventId),
     createdBy: (parent) => db.user.findByPk(parent.createdBy),
```
petorsong, line 33: same deal, let's validate for all three roles here

resolvers/patient.js
```diff
@@ -1,16 +1,34 @@
...
+        validators.demoRole
+      );
+      return db.patient.findAll({
         where: { collectionPointId: args.collectionPointId },
-      }),
+      });
+    },
   },
   Patient: {
     collectionPointId: (parent) =>
```
petorsong, line 34: validate for all 3 roles for these ones as well

resolvers/patient.js
```diff
@@ -44,6 +66,10 @@ const patientResolvers = {
       });
     },
     updatePatient: async (parent, args) => {
+      validators.validateRole(
+        [Roles.COMMANDER, Roles.SUPERVISOR],
```
petorsong, line 70: we'll need dispatch here since they'll be able to edit run numbers - limiting their ability to only edit that field will be done in the front end
```suggestion
        [Roles.COMMANDER, Roles.SUPERVISOR, Roles.DISPATCH],
```

petorsong (changes requested):
- in general, do we guarantee that the validator will go throw the error before the rest of the mutation if we don't define async/await?

- in general, we want all queries to be available to all roles as users of any role will be making those queries; in the requirements excel sheet, when we limit the viewing of any specific entity to specific roles, it mainly means that only those roles are allowed to go to the screens and explicitly see them (e.g. only commanders can go to the screen with all the users on it so they can create/edit/delete them, but everyone should be able to use the backend `users` query for unrelated queries)

- honestly this is an uber janky way to do enums, but it'll be a lot more pain for a lot less return if we set it up properly (so you're fine lol); proper way to do it is to treat the db as the source of truth - on page/app load, read the roles from the db and convert them into an appropriately-structured enum, then use that dynamically generated enum instead of statically defining it

resolvers/ambulance.js
```diff
@@ -1,34 +1,48 @@
...
+const { Roles } = require('../utils/enum');
 const validators = require('../utils/validators');
 
 const ambulanceResolvers = {
   Query: {
-    ambulances: () =>
-      db.ambulance.findAll({
+    ambulances: () => {
+      validators.validateRole(
+        [Roles.COMMANDER, Roles.SUPERVISOR, Roles.DISPATCH],
```
JasonYG, line 11: Small nitpick - maybe change this to `Object.keys(Roles)` or `Object.values(Roles)`, so we wouldn't have to add more items to this array if we were to add more Roles in the future. 

This might be intentional though - maybe we only want these three roles. Just something to think about

resolvers/ambulance.js
```diff
@@ -1,34 +1,48 @@
...
           },
         ],
-      }),
-    ambulance: (parent, args) =>
-      db.ambulance.findByPk(args.id, {
+      });
+    },
+    ambulance: (parent, args) => {
+      validators.validateRole(
+        [Roles.COMMANDER, Roles.SUPERVISOR, Roles.DISPATCH],
```
JasonYG, line 24: same thing here

resolvers/event.js
```diff
@@ -304,8 +336,9 @@ const eventResolvers = {
       });
       return db.event.findByPk(args.id);
     },
-    deleteEvent: async (parent, args) =>
-      Promise.all([
+    deleteEvent: async (parent, args) => {
+      validators.validateRole([Roles.COMMANDER], validators.demoRole);
```
JasonYG, line 340: I noticed that you've copy pasted this validator to a lot of these resolvers

Do you think there's a way of adding the validator to all these resolvers without copy-pasting? I just feel like this might not be ideal in terms of maintainability. In ordinary REST we might have middleware we attach to each endpoint; is there an alternative here?

JasonYG:
Left some inline comments, but some of my comments boil down to if we're able to add in default permissions to all resolvers. So for example, is there a way to add in default validators to all, say, hospital resolvers? If so, I can see that reducing the amount of copy-pasted permissions in.

resolvers/ambulance.js
petorsong: this is a good point - for these ones where we essentially just validate that they have a role, doing `Object.values(Roles)` makes a lot of sense

resolvers/event.js
petorsong: middleware is an interesting option, but since roles can be different for particular endpoints, wouldn't we still have to handle those specific cases? i agree it would be cleaner not to copy paste this so many times, but i dunno if the benefits of introducing middleware at this stage would outweigh the complexity

one could argue that this is more maintainable in one sense, but yeah in general it ain't great lol

resolvers/ambulance.js
rickrm: Sounds good!

resolvers/event.js
petorsong: @JasonYG

petorsong: offline discussion - not worth investing in middleware at this point

resolvers/ambulance.js
```diff
@@ -1,34 +1,48 @@
...
+const { Roles } = require('../utils/enum');
 const validators = require('../utils/validators');
 
 const ambulanceResolvers = {
   Query: {
-    ambulances: () =>
-      db.ambulance.findAll({
+    ambulances: () => {
+      validators.validateRole(
+        Object.keys(Roles),
```
JasonYG, line 11: lol super small nitpick, but should this be Object.values()?

resolvers/event.js
```diff
@@ -1,12 +1,17 @@
...
+const { Roles } = require('../utils/enum');
 const validators = require('../utils/validators');
 
 const eventResolvers = {
   Query: {
-    events: () =>
-      db.event.findAll({
+    events: () => {
+      validators.validateRole(
+        Object.keys(Roles),
```
JasonYG, line 11: ^

resolvers/event.js
```diff
@@ -15,9 +20,14 @@ const eventResolvers = {
...
           },
         ],
-      }),
-    event: (parent, args) =>
-      db.event.findByPk(args.id, {
+      });
+    },
+    event: (parent, args) => {
+      validators.validateRole(
+        Object.keys(Roles),
```
JasonYG, line 27: ^

petorsong (approved):

JasonYG (approved):

patrick-du (approved):

PR #70, opened 2020-10-15
Update role names

sarafong:
[Notion ticket](https://www.notion.so/uwblueprintexecs/update-user-access-levels-1488ff68ce2a419a85c82d2f690b21d2)

## Changes
- Updated `role` and `user` models to establish associations between the two
- Added `role` queries to `user` resolver
- Created migration to change column in `users` table from `accessLevel` to `roleId`, which now uses the `roles` table
- Merge together with [frontend change](https://github.com/uwblueprint/paramedics-react/pull/45)
## Testing
1. Run migration and verify that `users` table has `accessLevel` column removed and is replaced with `roleId` column
2. `roleId` column should have correct mappings - 1 is Commander, 2 is Supervisor (no mapping for dispatch)
3. Verify that `roles` query works in GraphQL
4. Verify frontend tests on branch `update-user-roles` in `paramedics-react` from respective frontend PR

before opening PR
- [x] check notion ticket
- [x] run linter
- [x] go through file diff

filling out PR
- [x] descriptive title
- [x] fill out `Changes`
- [x] fill out `Testing`

after opening PR
- [x] move ticket to `In PR`
- [ ] make sure build passes

### DIFF
--- a/migrations/20201015005801-update-user-roles.js
+++ b/migrations/20201015005801-update-user-roles.js
@@ -6,65 +6,22 @@ module.exports = {
       return Promise.all([
         queryInterface.addColumn(
           'users',
-          'role',
+          'roleId',
           { type: Sequelize.DataTypes.INTEGER },
           { transaction: t }
         ),
         queryInterface.sequelize.query(
-          `UPDATE users SET "role" = 1 WHERE "accessLevel" = 'COMMANDER' OR "accessLevel" = 'ADMIN'`,
+          `UPDATE users SET "roleId" = 1 WHERE "accessLevel" = 'COMMANDER' OR "accessLevel" = 'ADMIN'`,
           { transaction: t }
         ),
-        queryInterface.sequelize.query(
-          `UPDATE users SET "role" = 2 WHERE "accessLevel" = 'SUPERVISOR'`,
+        queryInterface.sequelize.query( 
+          `UPDATE users SET "roleId" = 2 WHERE "accessLevel" = 'SUPERVISOR'`,
           { transaction: t }
         ),
-        // queryInterface.sequelize.query(
-        //   `UPDATE users SET "role" = 3 WHERE "accessLevel" = 'DISPATCH'`
-        // ,
-        // { transaction: t }
-        // ),
         queryInterface.removeColumn('users', 'accessLevel', {
           transaction: t,
-        }),
+        })
       ]);
-      //   return Promise(() => {
-      //     queryInterface.addColumn(
-      //       'users',
-      //       'role',
-      //       { type: Sequelize.DataTypes.INTEGER },
-      //       { transaction: t }
-      //     )
-      //   })
-      //   .then(() => {
-      //     return Promise(() => {
-      //       queryInterface.sequelize.query(
-      //         `UPDATE users SET "role" = 1 where "accessLevel" = "COMMANDER"`,
-      //       { transaction: t }
-      //       )})
-      //       .then(() => {
-      //       return Promise(() => {
-      //         queryInterface.sequelize.query(
-      //           `UPDATE users SET "role" = 2 where "accessLevel" = "SUPERVISOR"`
-      //         ,
-      //         { transaction: t }
-      //       )})
-      //         .then(() => {
-      //           return Promise(() => {
-      //             queryInterface.sequelize.query(
-      //               `UPDATE users SET "role" = 3 where "accessLevel" = "DISPATCH"`
-      //             ,
-      //             { transaction: t }
-      //           )})
-      //           .then(() => {
-      //             return Promise(() => {
-      //               queryInterface.removeColumn('users', 'accessLevel', {
-      //                 transaction: t,
-      //               })
-      //               });
-      //           });
-      //         });
-      //       });
-      //   });
     });
   },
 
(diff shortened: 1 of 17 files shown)


migrations/20201015005801-update-user-roles.js
```diff
@@ -0,0 +1,57 @@
+'use strict';
+
+module.exports = {
+  up: (queryInterface, Sequelize) => {
```
petorsong, line 4: we also need to drop the `accessLevel` enum from the `users` table as well - [there's an example of it here](https://github.com/uwblueprint/paramedics-web/blob/a47ee4e7c19d4fc575bd2c870d7a896f5888c531/migrations/20200128014549-create-user.js#L44)

(and don't forget to put it back in during the down migration)

(also you'll need to drop it last within the up migration and add it first within the down migration)

migrations/20201015005801-update-user-roles.js
```diff
@@ -0,0 +1,57 @@
...
+        queryInterface.sequelize.query(
+          `UPDATE users SET "accessLevel" = 'COMMANDER' WHERE "roleId" = 1`,
+          { transaction: t }
+        ),
+        queryInterface.sequelize.query(
+          `UPDATE users SET "accessLevel" = 'SUPERVISOR' WHERE "roleId" = 2`,
+          { transaction: t }
+        ),
+        queryInterface.sequelize.query(
+          `UPDATE users SET "accessLevel" = 'DISPATCH' WHERE "roleId" = 3`,
```
petorsong, line 48: we don't have the `DISPATCH` access level before so this wouldn't work - we can swap it to `SUPERVISOR` instead (not ideal but i think this is the best option)
```suggestion
          `UPDATE users SET "accessLevel" = 'SUPERVISOR' WHERE "roleId" = 3`,
```

petorsong (changes requested):
nice work! models/resolvers are good to go, just some changes to the migrations
- since we're already touching the `users` table + schema, can we also remove the `emergencyContact` & `password` fields? (can be part of another migration)

migrations/20201015005801-update-user-roles.js
```diff
@@ -0,0 +1,57 @@
...
+  up: (queryInterface, Sequelize) => {
+    return queryInterface.sequelize.transaction((t) => {
+      return Promise.all([
+        queryInterface.addColumn(
+          'users',
+          'roleId',
+          { type: Sequelize.DataTypes.INTEGER },
+          { transaction: t }
+        ),
+        queryInterface.sequelize.query(
```
aaronabraham311, line 13: I'm curious: is there no Sequelize functionality for this type of update? Is a raw SQL query absolutely necessary?

migrations/20201015005801-update-user-roles.js
aaronabraham311: Agree with this! You can check if you are deleting enums properly with [this query](https://stackoverflow.com/questions/9540681/list-postgres-enum-type).

aaronabraham311 (changes requested):
Great work! Everything was working on my side, just need to fix up the `accessLevel` enum business.

migrations/20201015005801-update-user-roles.js
```diff
@@ -0,0 +1,57 @@
+'use strict';
+
+module.exports = {
+  up: (queryInterface, Sequelize) => {
+    return queryInterface.sequelize.transaction((t) => {
+      return Promise.all([
```
aaronabraham311, line 6: @petorsong and I realized that we cannot use `Promise.all` for this particular change. When we put all these operations in one promise, it is not guaranteed that these queries will execute in the order that you have specified here. For example, it is possible that the `accessLevel` column is deleted before any of your update queries are executed. Since some of these queries are inter-related, you will need to use promise chaining as specified in the code changes in #62.

sarafong: yep I will fix this. I set up promise chaining but it got really messy and I decided to be lazy and try `Promise.all` anyway. it somehow executes properly but def not a good way to do it

migrations/20201015005801-update-user-roles.js
sarafong: well there shouldn't be any data with `roleId=3` so this line probably won't run, it's more just to be explicit. `roleId=2` is already `SUPERVISOR`, so don't think it's correct to map those

migrations/20201015005801-update-user-roles.js
sarafong: hmm i'm not too familiar with all the functions but I copied this format from another previous migration that I think peter did haha

https://sequelize.org/master/class/lib/dialects/abstract/query-interface.js~QueryInterface.html from this list, it doesn't look like any of the functions can really achieve the same thing (other than maybe `bulkUpdate`? but I don't think that's what it does)

migrations/20201015005801-update-user-roles.js
sarafong: ah ok I forgot about that! will add that in

migrations/20201015005801-update-user-roles.js
aaronabraham311: Actually I ran into this issue when I was doing my maps ticket. I accidentally used your backend and tried to migrate down, but I had `DISPATCH` in my access levels.

sarafong: right - what exactly is the issue with mapping `DISPATCH` to `roleId = 3` in the down migration though? If I add a mapping for `roleId=3` => `DISPATCH` in the up migration everything should be fine?

migrations/20201015005801-update-user-roles.js
```diff
@@ -1,57 +1,57 @@
...
-          `UPDATE users SET "roleId" = 2 WHERE "accessLevel" = 'SUPERVISOR'`,
-          { transaction: t }
-        ),
-        queryInterface.removeColumn('users', 'accessLevel', {
-          transaction: t,
-        }),
-      ]);
+  up: async (queryInterface, Sequelize) => {
+    return queryInterface.sequelize.transaction(async (t) => {
+      await queryInterface.addColumn(
```
aaronabraham311, line 6: Rather than having multiple awaits, shouldn't we be using `.then()` syntax? This guarantees that the initial promise finishes before moving on to the next one.

petorsong: i think you might be a bit confused - both `awaits` and `.then()` waits for the initial promise to resolve before proceeding

there is a case to be made to group some of the awaits to run synchronously if they don't need to be done sequentially, but the performance gain is relatively minor and it's not particularly important

petorsong (approved):

aaronabraham311 (approved):

PR #73, opened 2020-11-04
Add basic auth functionality

JasonYG:
[Notion ticket](https://www.notion.so/uwblueprintexecs/396370c5e09f4c28bc7f455a7e84c908?v=3751aa62a2594c74b7e0ecb363304c24&p=5617c37a11884e02b85d405123c905d7)

## Changes
- Added `/login` and `/logout` routes that implement login and logout functionality with auth0 respectively
- Fixed issue with our localhost endpoint changing to `/graphql` from `/`
- Added the logged-in user's `User` object to the request session
- **Added new env variable** `AUTH0_CLIENT_ID` - not having this might break your backend! 
- See the notion ticket to see the new env variable

## Screenshots
N/A

## Testing
- Manually hit the login endpoint (`localhost:4000/login`) and try logging in using [these](https://www.notion.so/uwblueprintexecs/396370c5e09f4c28bc7f455a7e84c908?v=3751aa62a2594c74b7e0ecb363304c24&p=2c830c195e5f41e1980883be3cc0990f) credentials
- Manually hit the logout endpoint (`localhost:4000/logout`) after logging in
- Make sure that you're brought back to the auth0 screen after logging out, and trying to log back in
- Test the frontend to make sure we didn't introduce breaking changes - e.g. adding events, deleting patients, editing events, etc.

before opening PR
- [x] check notion ticket
- [x] run linter
- [x] go through file diff

filling out PR
- [x] descriptive title
- [x] fill out `Changes`
- [x] fill out `Testing`

after opening PR
- [x] move ticket to `In PR`
- [x] make sure build passes


certs/cert.pem
```diff
@@ -0,0 +1,34 @@
+-----BEGIN CERTIFICATE-----
```
JasonYG, line 1: @petorsong not sure if this file should be committed - it was already in the git history lol

Can you confirm if I should or shouldn't remove this?

certs/idp_key.pem
```diff
@@ -0,0 +1 @@
+MIIDXTCCAkWgAwIBAgIJALmVVuDWu4NYMA0GCSqGSIb3DQEBCwUAMEUxCzAJBgNVBAYTAkFVMRMwEQYDVQQIDApTb21lLVN0YXRlMSEwHwYDVQQKDBhJbnRlcm5ldCBXaWRnaXRzIFB0eSBMdGQwHhcNMTYxMjMxMTQzNDQ3WhcNNDgwNjI1MTQzNDQ3WjBFMQswCQYDVQQGEwJBVTETMBEGA1UECAwKU29tZS1TdGF0ZTEhMB8GA1UECgwYSW50ZXJuZXQgV2lkZ2l0cyBQdHkgTHRkMIIBIjANBgkqhkiG9w0BAQEFAAOCAQ8AMIIBCgKCAQEAzUCFozgNb1h1M0jzNRSCjhOBnR+uVbVpaWfXYIR+AhWDdEe5ryY+CgavOg8bfLybyzFdehlYdDRgkedEB/GjG8aJw06l0qF4jDOAw0kEygWCu2mcH7XOxRt+YAH3TVHa/Hu1W3WjzkobqqqLQ8gkKWWM27fOgAZ6GieaJBN6VBSMMcPey3HWLBmc+TYJmv1dbaO2jHhKh8pfKw0W12VM8P1PIO8gv4Phu/uuJYieBWKixBEyy0lHjyixYFCR12xdh4CA47q958ZRGnnDUGFVE1QhgRacJCOZ9bd5t9mr8KLaVBYTCJo5ERE8jymab5dPqe5qKfJsCZiqWglbjUo9twIDAQABo1AwTjAdBgNVHQ4EFgQUxpuwcs/CYQOyui+r1G+3KxBNhxkwHwYDVR0jBBgwFoAUxpuwcs/CYQOyui+r1G+3KxBNhxkwDAYDVR0TBAUwAwEB/zANBgkqhkiG9w0BAQsFAAOCAQEAAiWUKs/2x/viNCKi3Y6blEuCtAGhzOOZ9EjrvJ8+COH3Rag3tVBWrcBZ3/uhhPq5gy9lqw4OkvEws99/5jFsX1FJ6MKBgqfuy7yh5s1YfM0ANHYczMmYpZeAcQf2CGAaVfwTTfSlzNLsF2lW/ly7yapFzlYSJLGoVE+OHEu8g5SlNACUEfkXw+5Eghh+KzlIN7R6Q7r2ixWNFBC/jWf7NKUfJyX8qIG5md1YUeT6GBW9Bm2/1/RiO24JTaYlfLdKK9TYb8sG5B+OLab2DImG99CJ25RkAcSobWNF5zD0O6lgOo3cEdB/ksCq3hmtlC/DlLZ/D8CJ+7VuZnS1rR2naQ==
```
JasonYG, line 1: same here^

certs/key.pem
```diff
@@ -0,0 +1,52 @@
+-----BEGIN PRIVATE KEY-----
```
JasonYG, line 1: here too

server.js
```diff
@@ -1,11 +1,135 @@
...
+const samlStrategy = new saml.Strategy(
+  {
+    path: '/login/callback',
+    entryPoint:
+      `https://uwbp-paramedics.us.auth0.com/samlp/${process.env.AUTH0_CLIENT_ID}`,
+    acceptedClockSkewMs: -1, // SAML assertion not yet valid fix
+    logoutUrl: 'https://uwbp-paramedics.us.auth0.com/v2/logout',
+    additionalLogoutParams: {
+      // TODO: Move this to env variable
+      client_id: process.env.AUTH0_CLIENT_ID,
```
JasonYG, line 28: notice the new env variable for the auth0 `client_id`

server.js
```diff
@@ -1,11 +1,135 @@
...
+    logout: () => req.logout(),
+  }),
+  playground: {
+    settings: {
+      'request.credentials': 'same-origin',
+    },
+  },
+});
+
+server.applyMiddleware({ app, path: "/" });
```
JasonYG, line 131: @petorsong this `path: "/"` business fixes the localhost URL pointing to `/graphql` instead of `/`.

server.js
```diff
@@ -1,11 +1,137 @@
 'use strict';
 
-const { ApolloServer } = require('apollo-server');
+/* eslint-disable no-param-reassign */
```
JasonYG, line 3: we need this to stop eslint complaining about lines 47-49

server.js
```diff
@@ -1,11 +1,137 @@
...
+/* eslint-disable no-console */
```
JasonYG, line 4: added this for line 136

server.js
petorsong: can we wrap this around just the offending lines instead of the entire file?

server.js
petorsong: same thing for this

server.js
```diff
@@ -1,11 +1,137 @@
...
+const saml = require('passport-saml');
+const { ApolloServer } = require('apollo-server-express');
+const { Op } = require('sequelize');
+const db = require('./models');
 const { schema } = require('./graphql');
 
-const server = new ApolloServer({ schema });
+const PORT = 4000;
+const HOST = `http://localhost:${PORT}`;
+const SESSION_SECRET = 'notsecure';
```
petorsong, line 20: this is just for testing now right? if so can we add a todo on the line above to change this?

server.js
```diff
@@ -1,11 +1,137 @@
...
+
+const samlStrategy = new saml.Strategy(
+  {
+    path: '/login/callback',
+    entryPoint: `https://uwbp-paramedics.us.auth0.com/samlp/${process.env.AUTH0_CLIENT_ID}`,
+    acceptedClockSkewMs: -1, // SAML assertion not yet valid fix
+    logoutUrl: 'https://uwbp-paramedics.us.auth0.com/v2/logout',
+    additionalLogoutParams: {
+      // TODO: Move this to env variable
```
petorsong, line 29: right?
```suggestion
```

petorsong (changes requested):
login/logout works which is awesome! 🎉 could you add some additional comments on how things work? (at least for the major functions like `deserializeUser` and the strategy callback) it's currently hard for me to tell how the `User` object is being passed into every request and how to validate or test that fact

- anything new added to `paramedics-web.env` should be added to [this notion doc](https://www.notion.so/uwblueprintexecs/paramedics-web-env-348bb11f188e4d9fa99eebde2278224b) and you should ping everyone in `#paramedics-dev` once it gets merged in so it doesn't affect other devs

server.js
JasonYG: oh yeah good catch, will do

server.js
```diff
@@ -1,11 +1,148 @@
...
+        where: { email: { [Op.like]: email } },
+      })
+      .then((matchingUser) => {
+        // Save the nameId and nameIDFormat for logout
+        /* eslint-disable-next-line no-param-reassign */
+        matchingUser.saml = {};
+        /* eslint-disable-next-line no-param-reassign */
+        matchingUser.saml.nameID = profile.nameID;
+        /* eslint-disable-next-line no-param-reassign */
+        matchingUser.saml.nameIDFormat = profile.nameIDFormat;
```
petorsong, line 51: is there any particular reason we want to mutate `matchingUser` instead of assembling a separate object with the fields we care about? your comment on line 45 suggests that we only care about those two fields, but i'm guessing we have other fields we would also care about (e.g. email) - it would be better to be super explicit and list out exactly what we want and assemble that object manually instead of relying on looking at the `matchingUser` object and hoping that has everything we need

(this would also eliminate having to disable these linter rules - as a general rule of thumb, if you ever see that you're disabling rules like this, maybe reconsider what you're trying to do and if there's a better way to accomplish the same result)

server.js
```diff
@@ -1,11 +1,148 @@
...
+ *    it to req.user. After this operation, req.session.passport.user = user.email */ 
+passport.serializeUser((user, done) => {
+  done(null, user.email);
+});
+
+
+/** This is called with the email that was saved in serializeUser as the
+*     first argument to this function. (i.e. email = user.email). We would then
+*     use the user's email to get the actual user object (using db.user.findOne(...))
+*     Afterwards, the user object is attached to req.user */
```
petorsong, line 73: what's the difference between `req.session.passport.user` vs `req.user`?

server.js
```diff
@@ -1,11 +1,148 @@
...
+  if (!req.user) {
+    res.send('Login unsuccessful');
+  } else {
+    res.send('Login successful!');
+  }
+});
+
+app.get('/logout', (req, res) => {
+  if (req.user == null) {
+    res.send('No logged in user!');
```
petorsong, line 114: this will also be replaced right? if so please leave a comment

server.js
```diff
@@ -1,11 +1,148 @@
...
+    return;
+  }
+  samlStrategy.logout(req, function (err, uri) {
+    // Clear user login session on our end
+    res.clearCookie('connect.sid');
+    // Redirect to logout callback URL
+    return res.redirect(uri);
```
petorsong, line 121: where does `uri` come from?

server.js
```diff
@@ -1,11 +1,148 @@
...
+
+app.use(bodyParser.urlencoded({ extended: false }));
+app.use(bodyParser.json());
+
+app.get('/login', passport.authenticate('saml'));
+
+// TODO: Redirect to frontend auth landing page
+app.post('/login/callback', passport.authenticate('saml'), (req, res) => {
+  if (!req.user) {
+    res.send('Login unsuccessful');
```
petorsong, line 106: this will also be replaced right? if so please leave a comment

server.js
```diff
@@ -1,11 +1,148 @@
...
+  } else {
+    res.send('Login successful!');
```
petorsong, line 108: this will also be replaced right? if so please leave a comment

server.js
```diff
@@ -1,11 +1,148 @@
...
+  samlStrategy.logout(req, function (err, uri) {
+    // Clear user login session on our end
+    res.clearCookie('connect.sid');
+    // Redirect to logout callback URL
+    return res.redirect(uri);
+  });
+});
+// TODO: Redirect to frontend login screen
+app.get('/logout/callback', (req, res) => {
+  res.send('Logout callback');
```
petorsong, line 126: this will also be replaced right? if so please leave a comment

petorsong (changes requested):

server.js
```diff
@@ -1,11 +1,148 @@
...
+    logout: () => req.logout(),
+  }),
+  playground: {
+    settings: {
+      'request.credentials': 'same-origin',
+    },
+  },
+});
+
+server.applyMiddleware({ app, path: '/' });
```
JasonYG, line 143: this line changes the path back to `/` from `/graphql`

server.js
petorsong:
```suggestion
        const returnVal = {
          user: matchingUser.
          saml: {
            nameID: profile.nameID,
            nameIDFormat: profile.nameIDFormat,
          }
        }
```

JasonYG: good catch, made this change in the latest commit

server.js
JasonYG: so `req.session.passport.user` is used by passport to deserialize the user object, while `req.user` is the deserialized object. I'm guessing that `req.session.passport.user` includes extraneous metadata for `passport`.

server.js
JasonYG: good catch!

server.js
JasonYG: talked about this with you in call, but i'll leave a comment for clarity

petorsong (approved):
i think the only thing missing is adding the `User` object to every request, but i'm not entirely sure if that's included in this PR - my main concern is if this will affect any other functionality we currently have, if it doesn't i'm fine with merging

JasonYG:
> i think the only thing missing is adding the `User` object to every request, but i'm not entirely sure if that's included in this PR - my main concern is if this will affect any other functionality we currently have, if it doesn't i'm fine with merging

So just to clarify, by adding the `User` object to every request, do you mean adding it to all GraphQL queries + mutations?

Right now, the `User` object is attached to every HTTP request in `req.user` (this is done in deserializeUser), but I don't think I've dealt with queries and mutations.

I was thinking of adding a query later, something like `getCurrentUser` that returns the current logged-in user.

Thoughts?

petorsong:
every http request is fine - can you describe how i can test that though? i haven't been able to verify that at all

JasonYG:
> every http request is fine - can you describe how i can test that though? i haven't been able to verify that at all

yep for sure - so in any of our endpoints (e.g. `/login/callback`, `/logout/callback`, or you can even make a new route), try console.log'ing `req.user` after logging in. It should give you the proper user object

petorsong:
ok i'm satisfied that this PR shouldn't impact existing functionality too much so i'll merge

PR #77, opened 2020-11-16
Update hospitals and ambulances to be UNIQUE

patrick-du:
[Notion ticket](https://www.notion.so/uwblueprintexecs/no-duplicate-resources-8713dd5852f04c0eb67558cf562da051)

Associated frontend PR: https://github.com/uwblueprint/paramedics-react/pull/58

## Changes
- `migrations/20201116042500-update-ambulances-unique.js`
    - Remove all duplicate ambulances
    - Add UNIQUE constraint on `vehicleNumber` column of ambulance table

- `migrations/20201116042508-update-hospitals-unique.js`
    - Remove all duplicate hospitals
    - Add UNIQUE constraint on `name` column of hospital table

## Testing
- Create duplicate ambulances and hospitals
- Run migrations
- Check DB to ensure that duplicates no longer exist
- Try creating duplicate ambulances and hospitals again - this should not be possible
- Undo migrations and try creating duplicate ambulances and hospitals - this should be possible

before opening PR
- [x] check notion ticket
- [x] run linter
- [x] go through file diff

filling out PR
- [x] descriptive title
- [x] fill out `Changes`
- [x] fill out `Testing`

after opening PR
- [x] move ticket to `In PR`
- [x] make sure build passes

### DIFF
--- a/migrations/20201116042500-update-ambulances-unique.js
+++ b/migrations/20201116042500-update-ambulances-unique.js
@@ -1,0 +1,27 @@
+'use strict';
+
+module.exports = {
+  up: (queryInterface) => {
+    return queryInterface.sequelize.transaction(async (t) => {
+      await queryInterface.sequelize.query(
+        `DELETE FROM ambulances WHERE "id" NOT IN (SELECT * FROM (SELECT MIN("id") FROM ambulances GROUP BY "vehicleNumber") x)`,
+        { transaction: t }
+      );
+      queryInterface.addConstraint(
+        'ambulances',
+        {
+          fields: ['vehicleNumber'],
+          type: 'unique',
+        },
+        { transaction: t }
+      );
+    });
+  },
+
+  down: (queryInterface) => {
+    return queryInterface.removeConstraint(
+      'ambulances',
+      'ambulances_vehicleNumber_uk'
+    );
+  },
+};
--- a/migrations/20201116042508-update-hospitals-unique.js
+++ b/migrations/20201116042508-update-hospitals-unique.js
@@ -1,0 +1,24 @@
+'use strict';
+
+module.exports = {
+  up: (queryInterface) => {
+    return queryInterface.sequelize.transaction(async (t) => {
+      await queryInterface.sequelize.query(
+        `DELETE FROM hospitals WHERE "id" NOT IN (SELECT * FROM (SELECT MIN("id") FROM hospitals GROUP BY "name") x)`,
+        { transaction: t }
+      );
+      queryInterface.addConstraint(
+        'hospitals',
+        {
+          fields: ['name'],
+          type: 'unique',
+        },
+        { transaction: t }
+      );
+    });
+  },
+
+  down: (queryInterface) => {
+    return queryInterface.removeConstraint('hospitals', 'hospitals_name_uk');
+  },
+};


migrations/20201116042500-update-ambulances-unique.js
```diff
@@ -0,0 +1,27 @@
+'use strict';
+
+module.exports = {
+  up: (queryInterface) => {
+    return queryInterface.sequelize.transaction(async (t) => {
+      await queryInterface.sequelize.query(
+        `DELETE FROM ambulances WHERE "id" NOT IN (SELECT * FROM (SELECT MIN("id") FROM ambulances GROUP BY "vehicleNumber") x)`,
```
aaronabraham311, line 7: Just want to make sure I understand this query: you are grouping a database by vehicle numbers and then finding the minimum ID within each group of vehicle numbers. If the ambulances are not unique, we will be missing a particular ambulance ID, which indicates a duplicate?

migrations/20201116042500-update-ambulances-unique.js
```diff
@@ -0,0 +1,27 @@
...
+        },
+        { transaction: t }
+      );
+    });
+  },
+
+  down: (queryInterface) => {
+    return queryInterface.removeConstraint(
+      'ambulances',
+      'ambulances_vehicleNumber_uk'
```
aaronabraham311, line 24: Where do we define this constraint name?

aaronabraham311 (commented):
Great work! I tried adding a duplicate entity and I got the following result:
Screen Shot 2020-11-19 at 12 52 53 PM
Is this expected? Should we try to be more specific with this error?

migrations/20201116042500-update-ambulances-unique.js
patrick-du: Yeah, you got it. The ambulance table is grouped into rows that have the same `vehicleNumber` and the minimum ID is selected from each row. At this point, we have a table of the ambulances that are not duplicates. Finally, we delete all ambulances whose IDs are not in this table (duplicates).

After discussing with @petorsong, we decided to only keep the lowest ID of a duplicate ambulance/hospital. 

An alternative could have been to check the associations between the hospital/ambulances and other entities before deleting them - however, I believe (?) that we decided that there would be no clean approach to deleting duplicates and went with this option.

migrations/20201116042500-update-ambulances-unique.js
patrick-du: Sequelize automatically creates a name for it. Would it make more sense to manually name it for clarity and code readability purposes?

https://sequelize.org/master/class/lib/dialects/abstract/query-interface.js~QueryInterface.html#instance-method-addConstraint
- In the `options.name` row

patrick-du:
> Great work! I tried adding a duplicate entity and I got the following result:
> Screen Shot 2020-11-19 at 12 52 53 PM
> Is this expected? Should we try to be more specific with this error?

Yeah, that's expected. The error is shown in a bit more detail lower down. If that's not sufficient, I'm assuming there's a way we could explicitly have the first message field say "Cannot create duplicate hospitals/ambulances" or something along the lines of that but I'm not too familiar with GraphQL and overwriting their generated error messages.

image

migrations/20201116042500-update-ambulances-unique.js
aaronabraham311: I don't think it really matters, I was just curious.

aaronabraham311 (approved):

migrations/20201116042500-update-ambulances-unique.js
petorsong: yeah basically we're hoping that this shouldn't affect production/staging since there's not much data on those environments and even less complex association hierarchies - if it breaks on dev envs i'll just help whoever fix it

petorsong (approved):
🕋

PR #81, opened 2020-11-27
Support dependency injection for SAML strategies

axyjo:
[Notion ticket](https://www.notion.so/uwblueprintexecs/9224caeb21ce4350b8a1dc378d393323?v=43f44ac04d734394b7f7edc636c523b4)

## Changes
The general idea here is that I should be able to do something like when deploying to staging/production

```javascript
const application = require('./server');
application(adfsStrategy, '/graphql').listen(...production_things...)
```

## Screenshots
N/A

## Testing
- Please help! :)I don't have a local dev env for this project set up, so I'm relying on you, yes you reading this thing, to test this out.

before opening PR
- [ ] check notion ticket
- [ ] run linter
- [ ] go through file diff

filling out PR
- [ ] descriptive title
- [ ] update `notion ticket` link
- [ ] fill out `Changes`
  - [ ] \(optional) add inline comments in diff
- [ ] fill out `Testing`
- [ ] \(optional) add `Screenshots`
- [ ] assign yourself to the PR

after opening PR
- [ ] link PR to notion ticket
- [ ] move ticket to `In PR`
- [ ] make sure build passes
- [ ] ping `@ps` in `#paramedics-dev`

PR #82, opened 2020-11-29
Support multiple SAML strategies

petorsong:
## Changes
- adding support for multiple SAML strategies on passport via dependency injection
- functionally the same changes as #81

### DIFF
--- a/app.js
+++ b/app.js
@@ -1,0 +1,99 @@
+const express = require('express');
+const bodyParser = require('body-parser');
+const session = require('express-session');
+const passport = require('passport');
+const { ApolloServer } = require('apollo-server-express');
+const { Op } = require('sequelize');
+
+const db = require('./models');
+const { schema } = require('./graphql');
+
+function application({ authStrategy, graphqlPath, sessionSecret }) {
+  passport.use(authStrategy);
+
+  /** This saves user.email in the session, under req.session.passport.user,
+   *    which is later used by deserializeUser to find the user object and attach
+   *    it to req.user. After this operation, req.session.passport.user = user.email */
+  passport.serializeUser(({ userEmail }, done) => {
+    done(null, userEmail);
+  });
+
+  /** This is called with the email that was saved in serializeUser as the
+   *     first argument to this function. (i.e. email = user.email). We would then
+   *     use the user's email to get the actual user object (using db.user.findOne(...))
+   *     Afterwards, the user object is attached to req.user */
+  passport.deserializeUser(async (email, done) => {
+    done(
+      null,
+      await db.user.findOne({
+        where: { email: { [Op.like]: email } },
+      })
+    );
+  });
+
+  const app = express();
+
+  app.use(
+    session({
+      secret: sessionSecret,
+      resave: false,
+      saveUninitialized: false,
+    })
+  );
+
+  app.use(passport.initialize({}));
+  app.use(passport.session({}));
+
+  app.use(bodyParser.urlencoded({ extended: false }));
+  app.use(bodyParser.json());
+
+  app.get('/login', passport.authenticate('saml'));
+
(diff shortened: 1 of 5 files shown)


resolvers/patient.js
```diff
@@ -17,43 +17,31 @@ const patientResolvers = {
       subscribe: withFilter(
         () => pubsub.asyncIterator([PATIENT_ADDED]),
         (payload, variables) => {
-          return (
-            payload.eventId ===
-            parseInt(variables.eventId)
-          );
+          return payload.eventId === parseInt(variables.eventId);
```
petorsong, line 20: picked up some prettier formatting changes that was missed from a merged PR

server.js
```diff
@@ -2,153 +2,22 @@
...
-    res.send('No logged in user!');
-    return;
-  }
-  samlStrategy.logout(req, function (err, uri) {
-    // Clear user login session on our end
-    res.clearCookie('connect.sid');
-    // Redirect to logout callback URL
-    // Note: this uri is specified in the `returnTo` field of `samlStrategy`'s `additionalLogoutParams`
-    return res.redirect(uri);
+if (require.main === module) {
```
JasonYG, line 12: just to make sure I'm understanding this right, `require.main` corresponds to the file that we call with `node file.js`, where `require.main` would equal `file.js`, correct?

JasonYG: i.e. it's the "entry point" into the app, the file we call with the command `node`?

JasonYG: if so, how come this if statement is needed? If it's the case that `require.main !== module`, i.e. we're calling a DIFFERENT file, `server.js` wouldn't be called at all right since it isn't imported anywhere?

server.js
```diff
@@ -2,153 +2,22 @@
...
-const passport = require('passport');
-const saml = require('passport-saml');
-const { ApolloServer } = require('apollo-server-express');
-const { Op } = require('sequelize');
-const db = require('./models');
-const { schema } = require('./graphql');
+const App = require('./app');
 
 const PORT = 4000;
 const HOST = `http://localhost:${PORT}`;
```
JasonYG, line 8: should this HOST be specified here? would it make more sense to specify it in `app.js`, since I see that the only place we're using this is to console log the `App ready on ${HOST}`

app.js
```diff
@@ -0,0 +1,99 @@
...
+const bodyParser = require('body-parser');
+const session = require('express-session');
+const passport = require('passport');
+const { ApolloServer } = require('apollo-server-express');
+const { Op } = require('sequelize');
+
+const db = require('./models');
+const { schema } = require('./graphql');
+
+function application({ authStrategy, graphqlPath, sessionSecret }) {
```
JasonYG, line 11: does it make sense to pass the session secret as a parameter to this function, or would it be better to specify it as an .env variable?

JasonYG: and also, how come we want to pass the `graphqlPath` as a parameter?

JasonYG (changes requested):
Looked over PR and looks mostly good, left some comments/questions though.

Thanks for taking the time to make this PR!

server.js
petorsong: i believe we want to make sure that the server is only initialized if we explicitly call `node server.js` as the entrypoint so we have full control over how it runs and what settings/environment it's running in

server.js
petorsong: we don't use `HOST` in app either so it's not necessary - to be honest it's not really doing anything and only handy for development to ensure that your server spun up; overall let's just leave it in

app.js
petorsong: i think the idea is to do all our env var reading in `server.js` and pass it in to app - honestly no preference one way or another

`graphqlPath` is being used as the route to apply the express middleware on - see line 94

PR #83, opened 2020-11-30
Adding backend support for different location pin types

aaronabraham311:
[Notion ticket](https://www.notion.so/uwblueprintexecs/396370c5e09f4c28bc7f455a7e84c908?v=3751aa62a2594c74b7e0ecb363304c24&p=380f161e725846019b6d5002e046ff7c)

## Changes
- Added two new columns:
    - `pinType`: This is an enum column which indicates if a pin is an event, CCP or an 'other' pin
    - `ccpId`: This is a column to indicate the associated CCP for a CCP pin
- New validation:
    - If `pinType = EVENT`, then we want to make sure that the event given for a pin is valid and that `ccpParentId` is not filled
    - If `pinType = CCP`, then we want to make sure that a valid CCP is given

## Screenshots
Screen Shot 2020-11-30 at 1 33 39 PM
Screen Shot 2020-11-30 at 1 33 54 PM

## Testing
For each of the following, make sure that the database is reflected properly
- Add multiple event, CCPs and other pins 
- Edit multiple event, CCPs and other pins
- Delete and restore multiple event, CCP and other pins

before opening PR
- [x] check notion ticket
- [x] run linter
- [x] go through file diff

filling out PR
- [x] descriptive title
- [x] fill out `Changes`
- [x] fill out `Testing`

after opening PR
- [x] move ticket to `In PR`
- [x] make sure build passes

### DIFF
--- a/migrations/20201125223742-add-pin-type-and-parent-id.js
+++ b/migrations/20201125223742-add-pin-type-and-parent-id.js
@@ -1,0 +1,30 @@
+'use strict';
+
+module.exports = {
+  up: (queryInterface, Sequelize) => {
+    return queryInterface.sequelize.transaction((t) => {
+      return Promise.all([
+        queryInterface.addColumn('locationPins', 'pinType', {
+          type: Sequelize.DataTypes.ENUM('EVENT', 'CCP', 'OTHER'),
+        }, {transaction: t}),
+        queryInterface.addColumn('locationPins', 'ccpParentId', {
+          type: Sequelize.INTEGER,
+        }, {transaction: t}),
+        queryInterface.addColumn('locationPins', 'eventParentId', {
+          type: Sequelize.INTEGER,
+        }, {transaction: t})
+      ])
+    })
+  },
+
+  down: (queryInterface) => {
+    return queryInterface.sequelize.transaction((t) => {
+      return Promise.all([
+        queryInterface.removeColumn('locationPins', 'eventParentId', {transaction: t}),
+        queryInterface.removeColumn('locationPins', 'ccpParentId', {transaction: t}),
+        queryInterface.removeColumn('locationPins', 'pinType', {transaction: t}),
+        queryInterface.sequelize.query('DROP TYPE IF EXISTS "enum_locationPins_pinType"', {transaction: t}),
+      ])
+    })
+  }
+};
--- a/migrations/20201125223742-add-pin-type-and-parent-id.js
+++ b/migrations/20201125223742-add-pin-type-and-parent-id.js
@@ -9,39 +9,37 @@ module.exports = {
           'pinType',
           {
             type: Sequelize.DataTypes.ENUM('EVENT', 'CCP', 'OTHER'),
+            defaultValue: 'OTHER',
           },
           { transaction: t }
         ),
         queryInterface.addColumn(
           'locationPins',
-          'ccpParentId',
+          'ccpId',
           {
             type: Sequelize.INTEGER,
           },
           { transaction: t }
         ),
-      ]).then(() => {
-        queryInterface.sequelize.query(
-          `UPDATE "locationPins" SET "pinType" = 'OTHER' WHERE "pinType" IS NULL`
-        );
-      });
+      ]);
     });
   },
 
   down: (queryInterface) => {
     return queryInterface.sequelize.transaction((t) => {
       return Promise.all([
-        queryInterface.removeColumn('locationPins', 'ccpParentId', {
+        queryInterface.removeColumn('locationPins', 'ccpId', {
           transaction: t,
         }),
         queryInterface.removeColumn('locationPins', 'pinType', {
           transaction: t,
         }),
+      ]).then(() =>
         queryInterface.sequelize.query(
           'DROP TYPE IF EXISTS "enum_locationPins_pinType"',
           { transaction: t }
-        ),
-      ]);
+        )
+      );
     });
   },
 };
--- a/migrations/20201125223742-add-pin-type-and-parent-id.js
+++ b/migrations/20201125223742-add-pin-type-and-parent-id.js
@@ -4,27 +4,44 @@ module.exports = {
   up: (queryInterface, Sequelize) => {
     return queryInterface.sequelize.transaction((t) => {
       return Promise.all([
-        queryInterface.addColumn('locationPins', 'pinType', {
-          type: Sequelize.DataTypes.ENUM('EVENT', 'CCP', 'OTHER'),
-        }, {transaction: t}),
-        queryInterface.addColumn('locationPins', 'ccpParentId', {
-          type: Sequelize.INTEGER,
-        }, {transaction: t}),
-        queryInterface.addColumn('locationPins', 'eventParentId', {
-          type: Sequelize.INTEGER,
-        }, {transaction: t})
-      ])
-    })
+        queryInterface.addColumn(
+          'locationPins',
+          'pinType',
+          {
+            type: Sequelize.DataTypes.ENUM('EVENT', 'CCP', 'OTHER'),
+          },
+          { transaction: t }
+        ),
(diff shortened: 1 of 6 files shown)


migrations/20201125223742-add-pin-type-and-parent-id.js
```diff
@@ -0,0 +1,47 @@
...
+          'locationPins',
+          'ccpParentId',
+          {
+            type: Sequelize.INTEGER,
+          },
+          { transaction: t }
+        ),
+      ]).then(() => {
+        queryInterface.sequelize.query(
+          `UPDATE "locationPins" SET "pinType" = 'OTHER' WHERE "pinType" IS NULL`
```
aaronabraham311, line 25: All existing pins will be initialized as `OTHER` when this migration is run

utils/validators.js
```diff
@@ -96,22 +96,40 @@ module.exports = {
       }
     });
   },
-  validateLocationPin: (
-    locationPinId,
-    checkParanoid = false,
-    errorMessage = 'Invalid location pin ID: ' + locationPinId
-  ) => {
+  validateLocationPin: (params) => {
```
aaronabraham311, line 99: Using a params object to make this much cleaner

utils/validators.js
```diff
@@ -96,22 +96,40 @@ module.exports = {
...
+      if (!params.eventId) {
+        throw new Error('An event pin must be created by a valid event');
+      }
+
+      if (params.ccpParentId) {
+        throw new Error('An event pin cannot be created by a CCP');
+      }
+    }
+
+    if (!params.newPin) {
```
aaronabraham311, line 122: `newPin` is only `true` if it is being added. In other words, this validation is only for existing pins (on updates/deletes/restores)

migrations/20201125223742-add-pin-type-and-parent-id.js
```diff
@@ -0,0 +1,47 @@
...
+        );
+      });
```
petorsong, line 27: minor nit - also i would double check the formatting (run it through prettier)
```suggestion
      ]).then(() =>
        queryInterface.sequelize.query(
          `UPDATE "locationPins" SET "pinType" = 'OTHER' WHERE "pinType" IS NULL`
        );
     );
```

migrations/20201125223742-add-pin-type-and-parent-id.js
```diff
@@ -0,0 +1,47 @@
...
+        queryInterface.removeColumn('locationPins', 'ccpParentId', {
+          transaction: t,
+        }),
+        queryInterface.removeColumn('locationPins', 'pinType', {
+          transaction: t,
+        }),
+        queryInterface.sequelize.query(
+          'DROP TYPE IF EXISTS "enum_locationPins_pinType"',
+          { transaction: t }
+        ),
```
petorsong, line 43: is doing this concurrently ok? would it be better to remove the enum after `pinType` is deleted since that column relies on the enum to exist

resolvers/locationPin.js
```diff
@@ -27,9 +27,11 @@ const locationPinResolvers = {
   LocationPin: {
     eventId: (parent) => {
       validators.validateRole(Object.values(Roles), validators.demoRole);
-
       return db.event.findByPk(parent.eventId);
     },
+    ccpParentId: (parent) => {
+      return db.collectionPoint.findByPk(parent.ccpParentId);
+    },
```
petorsong, line 34: don't recall how this works exactly but would we need a null check here? `ccpParentId` is the first foreign key we've introduced that ISN'T mandatory, so just want to make sure we're doing it properly

utils/validators.js
```diff
@@ -96,22 +96,40 @@ module.exports = {
...
+      return db.locationPins
+        .findByPk(params.locationPinId, options)
+        .then((locationPin) => {
+          if (!locationPin) {
+            throw new Error(params.errorMessage);
+          }
+        });
+    }
+
+    return true;
```
petorsong, line 132: don't really need to return true - we only care if the validation fails and an error is thrown
```suggestion
```

utils/validators.js
```diff
@@ -96,22 +96,40 @@ module.exports = {
...
+        throw new Error('An event pin cannot be created by a CCP');
+      }
+    }
+
+    if (!params.newPin) {
+      return db.locationPins
+        .findByPk(params.locationPinId, options)
+        .then((locationPin) => {
+          if (!locationPin) {
+            throw new Error(params.errorMessage);
```
petorsong, line 127: we only use `errorMessage` once right here but we pass it in as the same `'Invalid location pin ID: ' + args.id` string every time - we can just use the string here and not pass any `errorMessage` in at all

utils/validators.js
petorsong: this is unnecessary - we can just check if `params.id` exists if we want to know if this is an existing pin or not

resolvers/locationPin.js
```diff
@@ -40,13 +42,25 @@ const locationPinResolvers = {
         validators.demoRole
       );
       await validators.validateEvent(args.eventId);
+      if (args.ccpParentId) {
+        await validators.validateCollectionPoint(args.ccpParentId);
+      }
```
petorsong, line 47: doesn't this happen inside `validateLocationPin`? no need to do it twice
```suggestion
```

petorsong (changes requested):
🀄 looks pretty good overall, but i think it can use some cleanup
- should we just rename `ccpParentId` to `ccpId`? (we already have `eventId` which captures the same type of relationship imo)
- it would be way easier to pass the entire `args` object into `validateLocationPins` - we can just pick out what we care about and the fields will pretty much look the same every time anyways (so no need to pick out specific fields and pass it along)
- (optional) we can fold in the other validations that happen in the mutation resolvers into `validateLocationPins`, so in each mutation resolver we'd only be calling `validateLocationPins(args)` which cuts out a lot of bloat in that file; additionally, all of our extra validation happens conditionally (via null checks usually) and we check pretty much the same conditions in `validateLocationPins` already, so i think it's a logical jump to stuff it all in there

resolvers/locationPin.js
aaronabraham311: This section of the code is referring to the nested information that we display on GraphQL. We don't need to do a null check for CCP because we are allowing pins to have empty `ccpId` fields, as you can see in the GraphQL schema.

utils/validators.js
```diff
@@ -1,3 +1,5 @@
+/* eslint-disable consistent-return */
```
petorsong, line 1: how come we need this now? we previously had functions in here that didn't have a `return` (i.e. only threw errors) and this wasn't an issue before

migrations/20201125223742-add-pin-type-and-parent-id.js
```diff
@@ -9,39 +9,37 @@ module.exports = {
           'pinType',
           {
             type: Sequelize.DataTypes.ENUM('EVENT', 'CCP', 'OTHER'),
+            defaultValue: 'OTHER',
```
petorsong, line 12: this is pretty neat 📸

utils/validators.js
aaronabraham311: Every other function except `validateRole` had a `return` at the end of the function. Furthermore, these functions did not have any if statements wrapping the return. In our case, we don't really care what the function returns for `validateLocationPins` and we have wrapped it in an if statement, which the linter doesn't like. `validateRole` is the only function that doesn't return anything and hence the linter doesn't complain. You can check out more details [here](https://eslint.org/docs/rules/consistent-return). 

I don't see a huge problem anyways because we are never using our function returns, unless I am missing something. Or, we can just return a boolean because it doesn't impact anything.

petorsong (approved):
we weren't using the return values of the validators so i removed them; also refactored the way we populate the options object we pass into sequelize for validators

